PR #116, opened 2024-05-16
111 fix authentication

dorianjagusch:
Authentication is fixed !!

Changes:
- Fixed CRFS settings for dev
- Fixed CORS settings for dev
- Fixed COOKIE settings for dev
- Added credential permission and CSRFToken to state-changing requests (PUT, POST, DELETE)
- Added logout view on backend
- Added logout view on frontend
- Made login and logout services use request service abstract class
- cleaned up console logs.

There is some commented code for nginx and the backend for later SSL usage

TO FIX:
- [ ] Prevent visiting pages after login/out that should not be accesible anymore
- [ ] Production settings

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,9 +1,15 @@
 FROM nginx:1.25.5
 
 COPY ./nginx.conf /etc/nginx/nginx.conf
-
 COPY ./index.html /usr/share/nginx/html/index.html
 COPY ./static /usr/share/nginx/html/static/
+
+# RUN apt-get update && apt-get install -y openssl && \
+# 	mkdir -p /etc/nginx/ssl && \
+# 	openssl req -newkey rsa:2048 -x509 -sha256 -days 365 -nodes \
+# 	-out /etc/nginx/ssl/nginx.crt \
+# 	-keyout /etc/nginx/ssl/nginx.key \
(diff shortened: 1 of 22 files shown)


frontend/Dockerfile
```diff
@@ -1,10 +1,16 @@
 FROM nginx:1.25.5
 
 COPY ./nginx.conf /etc/nginx/nginx.conf
-
 COPY ./index.html /usr/share/nginx/html/index.html
 COPY ./static /usr/share/nginx/html/static/
 
+# RUN apt-get update && apt-get install -y openssl && \
```
sakarisalmi, line 7: good idea of writing this out already!

sakarisalmi (approved):
Very nice stuff Dorian, your suffering of the past couple days was well worth it!

frontend/static/js/services/loginService.js
```diff
@@ -1,25 +1,19 @@
...
-	return request.then((response) => {
-		if (response.ok) {
-			const newData = response.json();
-			return newData;
-		} else {
-			throw new Error(`${response.status}`);
-		}
-	})
+		}));
+	}
```
merituulie, line 16: here is where we should add those other api requests and throw notimplemented as they are not meant to be called yet right?

merituulie (approved):

PR #117, opened 2024-05-17
91 privacy policy html attach to link

hussojo:
added privacypolicy.html file to be served from frontend.

next up:
- proper text for file
- modifying register page link/text to be more informative

### DIFF
--- a/frontend/static/js/components/formComponents/registerForm.js
+++ b/frontend/static/js/components/formComponents/registerForm.js
@@ -14,13 +14,16 @@ export default class extends AForm {
 		registerButton.classList.add('primary-btn');
 		registerButton.textContent = 'Sign up';
 
-		const linkElement = document.createElement('a');
-		linkElement.id = 'privacyPolicy';
-		linkElement.href = '../../../privacypolicy.html';
-		linkElement.textContent = 'Privacy Policy';
-		linkElement.target = '_blank';
-		linkElement.rel = 'noopener noreferrer';
+		//const linkElement = document.createElement('text');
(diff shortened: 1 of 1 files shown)


azarSarikhani (approved):
looks super good!

frontend/privacypolicy.html
```diff
@@ -0,0 +1,57 @@
...
+    </section>
+
+    <section>
+        <h2>Limitation of Liability</h2>
+        <p>We shall not be liable for any special, indirect, incidental, or consequential damages, including without limitation, damages resulting from loss of use, data, or profits, arising out of the use, inability to use, or the results of use of the website or mobile app, any websites linked to it, or any materials posted on it.</p>
+    </section>
+
+    <section>
+        <h2>Governing Law</h2>
+        <p>These terms and conditions shall be governed by and construed in accordance with the laws of [Your Country]. Any dispute arising out of or in connection with these terms and conditions shall be subject to the exclusive jurisdiction of the courts of [Your Country].</p>
```
dorianjagusch, line 45: These [Your Country] should maybe already say Finland.

dorianjagusch (approved):
We talked about that it would be ok, to have the policy as a separate page as that would make our lives easy by forcing the user to click the link and open a separate page. Upon further thinking about it, I fear that that may break the SAP nature of the website. It is fine for now and the changes can go in (besides maybe changing the country). But we should probably discuss in more detail how we can handle it. Sorry for not thinking about it properly before....

dorianjagusch (approved):

merituulie:
@hussojo was there still something coming in here or is this ready to merge?

PR #118, opened 2024-05-20
48 serverside pong

azarSarikhani:
This PR will add the first iteration of the server side pong game logic.
The back end server will measure ball position based on moving angle and possibility of a collision and will send messages to the front end every 50 milliseconds. The back is also always listening on player movements. This branch is tested by connecting to ws://localhost:8080/pong/ and sending messages to it and receiving messages from it.


backend/pong/game_logic.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class PongGame:
+    def __init__(self):
+        # Initialize game state
+        self.player1_y = PLAYGROUND_HEIGHT // 2  # Initial position for player 1 (midpoint)
+        self.player2_x = PLAYGROUND_WIDTH - WALL_MARGIN
+        self.player1_x = WALL_MARGIN
+        self.player2_y = PLAYGROUND_HEIGHT // 2  # Initial position for player 2 (midpoint)
+        self.ball_x = PLAYGROUND_WIDTH - WALL_MARGIN - PLAYER_WIDTH   # Initial position for ball (x-coordinate)
```
dorianjagusch, line 13: Not for this PR but for posterity, the ball should probably always start in the center with a random, but weighted probability distribution for its direction, kinda like a double funnel.

backend/pong/game_logic.py
```diff
@@ -0,0 +1,113 @@
...
+        self.game_started = True
+    
+    # Update ball position based on angle
+    def update_ball_position(self):
+        if self.game_started:
+            self.ball_x += BALL_SPEED * math.cos(math.radians(self.ball_angle))
+            self.ball_y += BALL_SPEED * math.sin(math.radians(self.ball_angle))
+            self.check_collisions()  # Check for collisions
+
+    def check_collisions(self):
```
dorianjagusch, line 60: We might need some kind of tolerance value here to avoid the dudes actually intersecting, which is going to be a little ugly

backend/pong/game_logic.py
```diff
@@ -0,0 +1,113 @@
...
+            self.ball_angle = 360 - self.ball_angle
+            
+        # Check for collisions with vertical walls
+        if self.ball_x <= 0:
+            self.handle_wall_collision('left')
+        elif self.ball_x >= PLAYGROUND_WIDTH - BALL_WIDTH:
+            self.handle_wall_collision('right')
+
+
+    def handle_player_collision(self, player_y):
```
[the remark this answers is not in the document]
dorianjagusch: Also, this is not going to work for both players (for one, because the relative intersect is calculated compared to player1, but because there is no consideration of the ball direction, the system just gives output in one direction

dorianjagusch (approved):
Great job! I left some comments. They are mostly what we discussed yesterday in our meeting and are not crucial for this PR, but just for future reference of things to work on once we have visuals. 

I didn't know where to put it, but we all should discuss how many turns there should be in the game or whether that's adjustable and whether we have increasing ball speed as a match progresses so games won't run infinitely long. But all that once we have the visuals

backend/backend/asgi.py
```diff
@@ -8,9 +8,17 @@
 """
 
 import os
-
+from channels.routing import ProtocolTypeRouter, URLRouter
 from django.core.asgi import get_asgi_application
 
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'backend.settings')
+# Initialize Django ASGI application early to ensure the AppRegistry
+# is populated before importing code that may import ORM models.
```
merituulie, line 16: Im not a fan of commenting code, comments get obsolete and honestly a basic django call can be googled in 1 sec if one needs to understand what is happenging.

backend/backend/settings.py
```diff
@@ -83,8 +86,8 @@
     },
 ]
 
-WSGI_APPLICATION = 'backend.wsgi.application'
-
+#WSGI_APPLICATION = 'backend.wsgi.application'
```
merituulie, line 89: let remove commented code, please! This can be found from the version control if we find we should use wsgi instead, which I doubt.

backend/pong/constants.py
```diff
@@ -0,0 +1,13 @@
...
+PLAYGROUND_HEIGHT = 400
+MESSAGE_INTERVAL_SECONDS = 0.05  # 50 milliseconds
+PLAYER_WIDTH = 1
+PLAYER_HEIGHT = 40
+BALL_WIDTH = 1
+BALL_HEIGHT = 1
+WALL_MARGIN = 2
+BALL_SPEED = 5 
+MAX_BOUNCE_ANGLE = 90  # Bounce angle in degrees
+MIN_BOUNCE_ANGLE = -90
```
merituulie, line 13: Is this a django thing, isnt it more favorable to use radians instead of degrees for angles as radians is more accurate?

backend/pong/game_logic.py
```diff
@@ -0,0 +1,113 @@
+# pong/game_logic.py
+from .constants import *
+import math
+
+
+class PongGame:
+    def __init__(self):
+        # Initialize game state
+        self.player1_y = PLAYGROUND_HEIGHT // 2  # Initial position for player 1 (midpoint)
```
merituulie, line 9: Im thinking as it is pong, maybe player_left and player_right would be more descriptive than player1 and player2...

merituulie: When wanting to rename anything like functions or variables in VS Code you can right click a highlighted symbol and select 'Rename symbol' and VS Code will find all references to that symbol and rename them:
image

merituulie (approved):
Looks good! In general I am a little bit concerned of the scalability of this solution as all of the information is in nthe game_logic file and there is not real seperation within the code.. But if all of the logic to play the game is there already, its okay!

PR #119, opened 2024-05-20
55 add full friendspageflow back and front

merituulie:
Added functionality from friends page to get current accepted friendships and either accept or decline pending friend requests. When either button is pressed, we refresh the page. Im a little bit concerned if I'm recursing the html stuff but seemed fine when tested locally.

Used the try catch block with this one as we are doign multiple async calls and I wanted the getHTML result not depend on the promise.then thingie, let me know how it looks and feels!

### DIFF
--- a/backend/Friends/models.py
+++ b/backend/Friends/models.py
@@ -4,14 +4,9 @@ from User.models import User
 from .managers import FriendsManager
 
 # Create your models here.
-class Friends(models.Model):
-	# User who initiated the friend request
-	user_id = models.ForeignKey(User, related_name='user_id', null=False, blank=False, on_delete=models.CASCADE)
-
-	# User who received the friend request
-	friend_id = models.ForeignKey(User, related_name='friend_id', null=False, blank=False, on_delete=models.CASCADE)
-
-	# Date and time when the friend request is sent
+class Friend(models.Model):
+	user_id = models.IntegerField()
+	friend_id = models.IntegerField()
 	insertTS = models.DateTimeField(auto_now_add=True)
 
 	objects = FriendsManager()
@@ -22,8 +17,6 @@ class Friends(models.Model):
 	class Meta:
 		# required because of RunTimeError that complains about lack of app_label
 		app_label = 'Friends'
-
-		# prevents a user from having the same relation to another user twice in the table
 		constraints = [
 			models.UniqueConstraint(fields=['user_id', 'friend_id'], name='unique_user_friend_pair')
 		]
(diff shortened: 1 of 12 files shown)


backend/Friends/models.py
```diff
@@ -4,14 +4,9 @@
...
-	# User who initiated the friend request
-	user_id = models.ForeignKey(User, related_name='user_id', null=False, blank=False, on_delete=models.CASCADE)
-
-	# User who received the friend request
-	friend_id = models.ForeignKey(User, related_name='friend_id', null=False, blank=False, on_delete=models.CASCADE)
-
-	# Date and time when the friend request is sent
+class Friend(models.Model):
+	user_id = models.IntegerField()
+	friend_id = models.IntegerField()
```
merituulie, line 9: ForeingKey made it so that we actually need to get the whole model of User and Friend through the Rest API which to me didnt seem liek a good idea, so changed these to integer fields. We need to add the check for the relation between actual users from the database later but didnt do it in this PR.

dorianjagusch: Have you made it an issue yet, just so we don't forget it?

frontend/static/js/components/requestCard.js
```diff
@@ -1,32 +1,36 @@
-const requestOptions = () => {
+const requestOptions = (user, acceptRequest, declineRequest) => {
```
dorianjagusch, line 1: Instead of individually adding Eventlisteners to each card, we could work with delegated eventlisteners on the scrollers or the overall page (see index.js). Once we have remove eventlisteners in place, it would make live much easier, as we wouldn't need to loop through all elements individually. 

Doesn't have to be done now. If anyone's interested one can look into `delegated eventlisteners` and `event bubbling` :D

dorianjagusch (approved):

backend/backend/exceptions.py
hussojo, line 1: Is this empty file here on purpose?

backend/Friends/models.py
merituulie: I was supposed to yesterday but did not, here we go: https://github.com/dorianjagusch/ft_transcendence/issues/120

backend/backend/exceptions.py
merituulie: Damnit this was left here, no its not! Ill remove it!

dorianjagusch (approved):

hussojo:
Nice work!!

hussojo (approved):

PR #123, opened 2024-05-24
122; removed pong/routing.py

sakarisalmi:
Added the stuff that was in pong/routing.py to backend/urls.py; modified asgi.py slightly; deleted pong/routing.py.
Built container, works.


azarSarikhani (approved):
You have made game_id a mandatory part of the url. I would like to know where it is coming from?

merituulie (approved):

sakarisalmi:
I wish I could reply to your message @azarSarikhani, but the match_id is coming from the Match and the Players I create in the database in the lobby when they send a POST request with the AuthenticatedGuestUserToken to the LaunchMatchView.
All of this will become a lot more clear when you see what I have in branch 115

PR #125, opened 2024-05-24
124 created Tokens app

sakarisalmi:
This was something I was originally going to have in the match-api pr, but I realised that it's most likely better to have it's own app and pr for it.
As the name says, I've created a new Tokens app. The tokens will be utilised with matches and tournaments.

In the models.py I have an AbstractToken which will work as the parent for all the tokens we might need with our project. The most important field in this abstract model is the 'token', which has a unique uuid string value. I have also created an AuthorizedGuestUserToken model that of course inherits from the AbstractToken. It has a couple extra foreignkeyfields such as host_user and guest_user, which will be set when the guest user authenticates themself. I've created a serializer for this model that only has the token field in it.

The idea is that when the guest user successfully authenticates themself in host user's lobby, the view will return two serializers like this:
```
{
    "token": {
        "token": "3bbfa841-73a7-45ac-b3dc-55bd6f214c28"
    },
    "guest_user": {
        "id": 2,
        "username": "guest",
        "is_online": false
    }
}
```
the guest_user data will come in handy when showing their profile, but if you don't want the current response body, please inform me and I will do something else then.
the token will be passed later to a view when launching the match (working on that currently on the 115 branch).

I've also created a view for removing the guest user. You pass the token to this view, and it will simply set the token's is_active field to false, which prevents its use later on.

I've tested the views with postman and everything works. You can check out the rest of the pr yourself, but it's good that especially @dorianjagusch knows how I imagine authenticating the guest user works exactly.

### DIFF
--- a/backend/Tokens/admin.py
+++ b/backend/Tokens/admin.py
@@ -1,3 +1,0 @@
-from django.contrib import admin
-
-# Register your models here.
--- a/backend/Tokens/admin.py
+++ b/backend/Tokens/admin.py
@@ -1,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
--- a/backend/Tokens/managers.py
+++ b/backend/Tokens/managers.py
@@ -1,39 +1,39 @@
 from django.db import models
 from django.db import transaction
-from datetime import datetime
+from django.utils import timezone
+from datetime import timedelta
 
 class AuthenticatedGuestUserTokenManager(models.Manager):
 	def get_or_create_token(self, host_user, guest_user):
-		with transaction.atomic():
-			try:
-				guest_token = self.get_queryset().filter(
-                    host_user=host_user,
-                    guest_user=guest_user
-                ).latest('created_at')
+		try:
+			guest_token = self.get_queryset().filter(
+				host_user=host_user,
+				guest_user=guest_user
+			).latest('created_at')
 
-				if guest_token.is_expired():
-					# If the latest token has expired, deactivate it and create a new one
-					guest_token.is_active = False
-					guest_token.save()
-					guest_token = self.create(
-						host_user=host_user,
-						guest_user=guest_user
-					)
-				else:
-					# If the latest token has not expired and is still active, update creation time, else create a new token
-					if guest_token.is_active == True:
-						guest_token.created_at = datetime.now() 
-						guest_token.save()
-					else:
-						guest_token = self.create(
-						host_user=host_user,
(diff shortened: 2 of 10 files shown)


backend/Tokens/admin.py
merituulie, line 1: I dont think we use the admin.pys so should be removed!

backend/Tokens/tests.py
merituulie, line 1: maybe would be good to add tests, if you want to dabble with that. mostly the view / managers as they have actual logic there.

backend/Tokens/managers.py
```diff
@@ -0,0 +1,39 @@
+from django.db import models
+from django.db import transaction
+from datetime import datetime
+
+class AuthenticatedGuestUserTokenManager(models.Manager):
+	def get_or_create_token(self, host_user, guest_user):
+		with transaction.atomic():
```
merituulie, line 7: Nice! In general, transaction scope should be as small as possible and only there when needed - aka there are at least two separate but co-dependent database / state changing calls to a separate system. As everything that can go wrong within a transaction scope will revert the changes, the bigger the transaction, more exceptions that will revert the changes. Not all exceptions / error cases necessarily mean that the db calls need to reverted to recover from the error. For the scalability I would:
1. Move this if else logic and transactional stuff to a manager.py and
2. Make sure the transaction scope is only wrapping the necessary parts - only inside the if statements that actually do 2 operations

merituulie (approved):

merituulie:
So, Django provides a way to generate the uuid just by determing such a token?

backend/Tokens/admin.py
sakarisalmi: Ok

backend/Tokens/tests.py
sakarisalmi: Ok, I'll do it in a later pr, but I don't know actually know how testing functions work in Python or other languages either, so it'll take a bit.

backend/Tokens/managers.py
sakarisalmi: ok to the second point, but what do you mean with the first? This is a managers.py file.

sakarisalmi:
Answering to @merituulie: seems so, but I'll read more on it.

backend/Tokens/managers.py
sakarisalmi: I removed the whole atomic transaction from the function as it didn't require it here.

dorianjagusch (approved):

PR #129, opened 2024-05-27
48 serverside pong

azarSarikhani:
This PR adds suggestions made to the previous game logic PR.


backend/pong/game_logic.py
```diff
@@ -27,25 +29,25 @@ def update_player_positions(self, key_press):
...
-            self.player2_y -= PLAYER_MOVEMENT_UNIT  # Move player 2 up
+            self.right_player_y -= PLAYER_MOVEMENT_UNIT
         elif key_press == 'l':
-            self.player2_y += PLAYER_MOVEMENT_UNIT  # Move player 2 down
+            self.right_player_y += PLAYER_MOVEMENT_UNIT
         
         # Ensure player positions stay within the bounds of the play area
-        self.player1_y = max(min(self.player1_y, PLAYGROUND_HEIGHT - PLAYER_HEIGHT), 0)  # Bound player 1 between 0 and (playground height - player height)
-        self.player2_y = max(min(self.player2_y, PLAYGROUND_HEIGHT - PLAYER_HEIGHT), 0)  # Bound player 2 between 0 and (playground height - player height)
+        self.left_player_y = max(min(self.left_player_y, ( PLAYGROUND_HEIGHT / 2 ) - PLAYER_HEIGHT), 0)  # Bound lfet side player between 0 and (playground height / 2  - player height)
```
dorianjagusch, line 42: typo :P

backend/pong/game_logic.py
```diff
@@ -77,37 +79,44 @@ def check_collisions(self):
...
+            self.loser = 'right_player'
+        elif self.right_player_score == 5:
+            self.game_over = True
+            self.winner = 'right_player'
+            self.loser = 'left_player'
+
 
     def get_game_state(self):
-        # Construct JSON message with player and ball positions
         game_state = {
```
dorianjagusch, line 104: I somehow overlooked this part in the previous review 😅 Would maybe be good to have a bit more structure too the json eg:

 game_state = {
            'ball': {
                'position': {
                    'x': self.ball_x,
                    'y': self.ball_y
                },
                'size': BALL_WIDTH,   # I assume the ball is round/square
            },
            'players': {
                  'size': {.   #unless we go for games customisation, the sizes will be the same, so no reason to duplicate it
                      'width': PLAYER_WIDTH,
                      'height': PLAYER_HEIGHT
                  },
                'left': {
                    'position': {
                        'x': self.left_player_x,
                        'y': self.left_player_y
                    },
                    'score': self.left_player_score
                },
                'right': {
                    'position': {
                        'x': self.right_player_x,
                        'y': self.right_player_y
                    },
                    'score': self.right_player_score
                }
            },
            'game': {
                'wall_margin': WALL_MARGIN,
                'over': self.game_over,
                'winner': self.winner,
                'loser': self.loser
            }
        }
        self.send(text_data=json.dumps(game_state))

dorianjagusch:
Noice! I am wondering whether we can cut down on some of the data being sent with every data exchange. Like player width/height/wall margin. Things that don't constantly change. But we can do that once the frontend is working. no reason to block here

dorianjagusch (approved):

backend/pong/game_logic.py
dorianjagusch: Sorry the indentation vanished...

merituulie: Oh true! Would probably make sense to make a struct / object with these values and parse into a json object and back!

merituulie: would make it easier to also then inject the ai instead of another player!

sakarisalmi (approved):
Nice!

PR #130, opened 2024-05-27
121 add flow to go into a users profile by clicking one of the friend page cards

merituulie:
Not pretty, but works. Here's what the flow looks like currently:

https://github.com/dorianjagusch/ft_transcendence/assets/44494010/cbd744d6-cefb-476d-a5ce-28b0d65a9089


frontend/static/js/components/profileComponents/buttonBar.js
```diff
@@ -12,8 +13,8 @@ const buttonBar = (buttons) => {
 	const buttonBar = document.createElement('div');
 	buttonBar.classList.add('button-bar');
 
-	buttons.forEach(({ className, textContent }) => {
-		const button = Button(className, textContent);
+	buttons.forEach(({ className, textContent, handler }) => {
+		let button = Button(className, textContent, handler);
```
dorianjagusch, line 17: Is there a reason this is let and not const? (besides the JS philosophical question of 'proper usage of cost')

dorianjagusch (approved):
Looks good. There are some minor things that are happening under the hood in JS, where we don't make the best use of memory (I have that in some of the pages as well, so whatever), but since it's just a hand full of functions, it doesn't really matter I would say.

Is there a specific reason you had to resort to `super` in the profilePage.js? Or is it for readability so it's clear that the functions are defined in a parent?

frontend/static/js/components/profileComponents/buttonBar.js
merituulie: no :D I mean I was first setting the handler to the button after the initial construction in Button, but left it in the Button constuction :D will fix

merituulie:
> Looks good. There are some minor things that are happening under the hood in JS, where we don't make the best use of memory (I have that in some of the pages as well, so whatever), but since it's just a hand full of functions, it doesn't really matter I would say.
> 
> Is there a specific reason you had to resort to `super` in the profilePage.js? Or is it for readability so it's clear that the functions are defined in a parent?

I think I thought it might have to do with the binding of the function and as that function is not available in the parent, we need to 
 refer to the super. Didn't really look into it (was just working through errors until something worked) actually as I didn't even remember doing that :D Ill figure it out next but lets go with this for now!

dorianjagusch (approved):

PR #132, opened 2024-05-28
131 persistent login

dorianjagusch:
Login navbar is fixed.

Changes:
- As long as the user has an active session on the frontend, they can only access the "internal" pages, but not the login/register and landing pages and vice versa. 
- Navigation via search bar also maintains the internal login state.
- Logging out and going back in the browser history sends you to the login page
- Moved navbar setting into a lone standing function


frontend/static/js/index.js
```diff
@@ -1,8 +1,7 @@
-import { router, navigateTo } from './router.js';
+import {router, navigateTo} from './router.js';
 import logoutService from './services/logoutService.js';
 
-
-sessionStorage.setItem('isLoggedIn', false);
+// localStorage.setItem('isLoggedIn', false);
```
merituulie, line 4: is this by design commented out?

sakarisalmi (approved):
Looks good to me

frontend/static/js/index.js
dorianjagusch: Yes and no. I did want it gone though to see whether I corrected everything that depended on the initialisation there (which nothing does anymore)

merituulie (approved):

PR #133, opened 2024-05-29
115 match api

sakarisalmi:
This is a major pr that adds the Tokens, Match, and Player apps to the backend. Now you are able to launch a single match as well as a test match, which was added for developing the graphics and gameplay of the pong game.

I recommend reading through the last couple of presentations in discord. I've refactored some stuff from the previous time I held a presentation, but you can read it yourself.

I've also made some postman tests for the Token and Match stuff, but I haven't fixed the previous tests that are currently failing because of new changes in the requests and responses.


backend/Match/serializers.py
```diff
@@ -0,0 +1,13 @@
...
+class MatchSerializer(serializers.ModelSerializer):
+	state_display = serializers.CharField(source='get_state_display', read_only=True)
+
+	
+	# tournament_id = serializers.PrimaryKeyRelatedField(queryset=Tournament.objects.all())
+	# tournament_phase_id = serializers.PrimaryKeyRelatedField(queryset=TournamentPhase.objects.all())
+
+	class Meta:
+		model = Match
+		fields = ['id', 'state', 'state_display', 'start_time', 'end_time', 'created_at', 'updated_at']
```
merituulie, line 13: insert_ts, updated_ts to follow the convention with previously used columns that present the same thing!

backend/Player/serializers.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import serializers
+from .models import Player
+from Match.models import Match
+from User.models import User
+
+class PlayerSerializer(serializers.ModelSerializer):
+	user = serializers.PrimaryKeyRelatedField(queryset=User.objects.all())
+	match = serializers.PrimaryKeyRelatedField(queryset=Match.objects.all())
```
merituulie, line 8: we use foreing keys elsewhere as user_id, match_id

backend/backend/urls.py
```diff
@@ -27,8 +29,11 @@
 from Friends.views import FriendsListView, \
 						FriendshipDetailView
 
-from Tokens.views import GuestUserAuthenticationView, \
-							DeactivateGuestUserTokenView
+from Tokens.views import SingleMatchGuestUserAuthenticationView, \
+							DeactivateMatchTokenView
+
+from Match.views import LaunchSingleMatchView, \
+							LaunchTestMatchView
```
merituulie, line 36: not sure I agree with these extra line breaks every where :D

backend/backend/urls.py
```diff
@@ -46,6 +51,14 @@
...
 
 	# Tokens views
-	path('authenticate-guest-user/', GuestUserAuthenticationView.as_view()),
-	path('remove-guest-user/', DeactivateGuestUserTokenView.as_view()),
+	path('authenticate-guest-user-for-single-match/', SingleMatchGuestUserAuthenticationView.as_view()),
+	path('deactivate-match-token/', DeactivateMatchTokenView.as_view()),
+
+	# Match Views
+	path('start-single-match/', LaunchSingleMatchView.as_view()),
+	path('start-test-match/', LaunchTestMatchView.as_view()),
```
merituulie, line 59: I think these routes are .... well weird :P To think about this Rest API we are building in a way that it would be build in a consistent way that thinking of a new / external client starting to use it, we should use the convention set previously. The new client should be able to understand what exactly does this route call do, what kind of data does it update / change?

Are these routes not for POSTing a match? or GETing a launched match? so wouldnt something like GET /match/ or POST /match/ or even POST /match/start would make it easier to dynamically also use the conventions we already have in the frontend too to use these routes?

backend/Match/managers.py
```diff
@@ -0,0 +1,73 @@
...
+from django.db.models import Q
+
+from .PlayerMatchStatus import PlayerMatchStatus
+from Player.models import Player
+from User.models import User
+
+import sys
+
+class MatchManager(models.Manager):
+	def create_match_and_its_players(self, left_side_user_id, right_side_user_id):
```
merituulie, line 11: As discussed in the meeting, I think the logic of creating the players and the match should be moved to some utility / shared manager where then we can import both User / Player and the Match. These two entities kind of depend on each other and we should decouple them to be able scale the business logic later. So, instead of the Player being dependent on a match or match being dependent on a player, they rather should both be dependent on some PlayerMatchManager!

backend/Match/serializers.py
sakarisalmi: ok, but wouldn't you rather change the current insert_ts-s and updated_ts-s into created_at and updated_at?

backend/Player/serializers.py
sakarisalmi: yes, but look at the latest presentation; I recommend we switch the previous user_ids etc. into simply User for the reasons specified in the presentation.

backend/backend/urls.py
sakarisalmi: I understand what you mean. Those are placeholders as I haven't started on the tournament api yet so I'm not sure how the endpoints for that will work yet.

I could see there being stuff like /tournament/authenticate-guest, in which case I could see single matches having stuff like /single-match/authenticate-guest and /single-match/start-match, or instead of single-match you'd have something like /lobby/etc.

backend/Match/managers.py
sakarisalmi: sure, sounds good

backend/Match/serializers.py
sakarisalmi: it's a bit weird that all the other fields don't have capitalised letters, but then we have something like insertTS for the User and Friends models XD

sakarisalmi:
I've created GameSetupManager into shared_utilities, which now has the create_match_and_its_players function. The function has also been refactored to take the MatchToken as a parameter.

azarSarikhani (approved):
This is quite too much for me, but I trust that you have tested it yourself.

backend/Match/serializers.py
merituulie: _ts aka timestamp, which tell us what kind of data we hold in the column, same with start_ts and end_ts or smt would be good, unless Django supports getting just the time

merituulie: In anycase I would indeed change either the existing data or these columns so that all columns that hold timestamps / datetime values for when the row was crreated / updated, will look the same. Its really for the other coders to know what this column does and we should really try to hold somekind of conventions. Think about it, if we have a column called 'name' in one table and then on another table something like 'called_by' or smt, would be confusing at least for me to understand that these both columns mean the name of the specific row.

backend/backend/urls.py
merituulie: this is basically the same issue as with the column names: we need to start having conventions on how and what we mean when we say this or that. One route to get users cant be /users/ and adding one POST /users/ and then updating a user would out of nowhere be PUT /update-user/. Please apply the existing convention! Here is what I suggested in the Discord:
_'can you change the token routes to follow the convetion of other routes? POSTin a token should be something like POST /tokens/ or /tokens/guest/ and the inactivation route should be PUT method and either the same as the other one, /tokens/ or /tokens/guest/'_

merituulie: for the match, it could be something like /match/?command=start for example, like the friends route

dorianjagusch: I am with Meri on that. Fully understandable that you had to navigate that in your own and did the best you knew how, but after some time away and given the time line we might want to step it up and set up things properly from the get go. I will approve the PR now and create an issue. I leave it up to you to fix it either now or on the new issue :D

dorianjagusch (approved):

backend/Match/serializers.py
sakarisalmi: ok, I'll make an issue and branch where I changed all the models' time fields so that they all have the FOO_ts format.

backend/backend/urls.py
sakarisalmi: I'll make a new issue for this!

backend/Match/PlayerMatchStatus.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
```
merituulie, line 1: Ill approve for now, but just a heads up the file name is not capitalized in the other enum file we have on the backend.

merituulie (approved):

merituulie (approved):

backend/Match/PlayerMatchStatus.py
sakarisalmi: I'll fix it in my next pr

PR #138, opened 2024-06-04
137 - refactored LaunchMatchView

sakarisalmi:
Refactored some stuff that were brought up in #133  review.

- Refactored LaunchSingleMatchView into MatchView
  - now it's a GET request that is passed the token value you get from authenticating the guest user like '{{baseurl}}/match/?token={{token_value}}'
- changed the file PlayerMatchStatus.py into playerMatchStatus.py
- updated the tests for match, token folder in postman collection (old version, not @merituulie new version)

### DIFF
--- a/backend/Match/views.py
+++ b/backend/Match/views.py
@@ -5,37 +5,35 @@ from rest_framework.request import Request
 from rest_framework import status
 from django.shortcuts import redirect
 
-from User.models import User
-from .models import Match
 from Tokens.models import MatchToken
-from Tokens.serializers import MatchTokenSerializer
 from shared_utilities.GameSetupManager import GameSetupManager
 from django.utils.decorators import method_decorator
-from shared_utilities.decorators import must_be_authenticated, \
-											valid_serializer_in_body
+from shared_utilities.decorators import must_be_authenticated
 
+class LaunchMatchView(APIView):
+    @method_decorator(must_be_authenticated)
+    def get(self, request):
+        token_str = request.query_params.get('token')
+        if not token_str:
+            return Response({'error': 'Token parameter is required'}, status=status.HTTP_400_BAD_REQUEST)
 
-class LaunchSingleMatchView(APIView):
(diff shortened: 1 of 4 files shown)


backend/Match/views.py
```diff
@@ -5,37 +5,35 @@
...
-		return redirect(pong_match_url) # this is prob not correct as this is a single-page app...
+        if not token.is_active or token.is_expired():
+            return Response({'error': 'Expired token'}, status=status.HTTP_403_FORBIDDEN)
+        if token.user_left_side.id != request.user.id:
+            return Response({'error': 'You are not the host user in the token'}, status=status.HTTP_403_FORBIDDEN)
+
+        match, player_left, player_right = GameSetupManager.create_match_and_its_players(token)
+
+        if not all([match, player_left, player_right]):
+            return Response({'error': 'Something went wrong when creating the match and players'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
merituulie, line 33: we dont really want to CHOOSE to return a 500 really - 500 means that we encountered an unhandled exception / exception we did not think of, rather 400 or smt..

backend/Match/views.py
```diff
@@ -5,37 +5,35 @@
...
+
+        pong_match_url = f'http://localhost:80/pong/{match.id}?token={token.token}'
```
merituulie, line 35: the url localhost etc part could be moved into a constant too, but maybe on the next branch? this would be annoying to find if / when we change the port / domain etc.

merituulie (approved):

backend/Match/views.py
sakarisalmi: That's true, and if we get a 500 in the eval that's a fail, but it also is a server side error if something goes wrong here

backend/Match/views.py
sakarisalmi: sure, sounds good.

PR #146, opened 2024-06-11
fix profileStats import

hussojo:
Sorry for super tiny pr, thought there was more to it :D

Problem was, we could not load any page, because profile stats didn't have export default anymore.


merituulie (approved):

dorianjagusch (approved):
Good. I fixed it on my branch, but better sooner than later. Thx

PR #147, opened 2024-06-13
Put back the user + friend foreing keys to ensure the relationships for Friend rows!

merituulie:
In Django, when you define a foreign key relationship, Django automatically creates a database column with the name of the field followed by _id. Therefore, even though your model fields are named user and friend, the corresponding database columns will be user_id and friend_id.


dorianjagusch (approved):
Do I understand correctly that you're changing the default name by Django to (with _id) to just the column name we want?

azarSarikhani (approved):

merituulie:
I'm adding the foreing key which in Django is the whole User row, not just the _id. Then the _id column comes by default with Django.

dorianjagusch (approved):

azarSarikhani (approved):

PR #149, opened 2024-06-15
91 privacy policy html attach to link

hussojo:
Finalized the privacypolicy.html with relevant information.
changed the element to use setAttribute to be consistent with other code.

First I had the html in a same styling as the rest of the app, but it didn't feel 'proper' or right, so made it plain and simple and more 'serious' looking, i u know what i mean :D

Not the most fancy shmancy, but it works.


frontend/privacypolicy.html
```diff
@@ -0,0 +1,92 @@
...
+        <li>To allow you to participate in interactive features of our Service</li>
+        <li>To track your game progress and achievements</li>
+        <li>To manage friendships within the game</li>
+    </ul>
+
+    <h2>3. Your Data Protection Rights</h2>
+
+    <p>Under the General Data Protection Regulation (GDPR), you have the following rights regarding your data:</p>
+    <ul>
+        <li><strong>The right to access</strong> – You have the right to request copies of your personal data.</li>
```
dorianjagusch, line 84: Does this mean that we need them to be able to get a table/ file of all their data? Similar as like instagram/snapchat etc?

dorianjagusch (approved):

frontend/privacypolicy.html
```diff
@@ -0,0 +1,92 @@
...
+	<meta charset="UTF-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<title>Privacy Policy</title>
+
+	<link rel="preconnect" href="https://fonts.googleapis.com">
+	<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+	<link href="https://fonts.googleapis.com/css2?family=Kode+Mono:wght@400..700&display=swap" rel="stylesheet">
+
+	<script type="module" src="./static/js/index.js" defer></script>
+	<link rel="stylesheet" href="./static/css/styles.css">
```
azarSarikhani, line 13: Do we need to keep these comments here? I am not familiar with how html pages conventionally are written, do they have a head an some references commented out?

azarSarikhani (approved):

frontend/privacypolicy.html
hussojo: Yep in theory. But in reality I think it's enough that we have thought about this, and have a way for the user to contact. So that we have taken in to account that GDPR is valid, but because we don't really save any data longer than our containers are up...

frontend/privacypolicy.html
hussojo: No need to keep, I thought I deleted those :D I'll make the change!

dorianjagusch (approved):

PR #152, opened 2024-06-21
Update requirements, change image name for back and front

merituulie:
- Added sudos to Makefile be sure make does not fail there.
- Updated packages in requirements.txt
- Changed the image tag name, if the compose is given a docker hub existing image:tag combination, it can pull the image from the docker hub, 42 is just too used as all of Hive uses this number.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ CYAN = \033[0;96m
 all: up
 
 up:
-	${DOCKER_COMPOSE} up --build --detach
+	sudo ${DOCKER_COMPOSE} up --build --detach
 	@echo "${GREEN}${NAME} is up!${C_RESET}"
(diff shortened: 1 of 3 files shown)


azarSarikhani (approved):
super!

dorianjagusch (approved):
Does the sudo addition work for people not on linux or wsl?
We had it before and it broke onn school macs. I removed it and instead have been invoking make with `sudo make`. 
I haven't been to school in a while, but can check it when the piscine has calmed down.

merituulie:
Yep lets check! I work with WSL so it does work indeed but as I felt the majority does work remotely its less work to not type the sudo everytime. In anycase in the end we should definitely test it during pre-eval

PR #153, opened 2024-06-23
139 add stats app

azarSarikhani:
This minimal PR will just add the skeleton of the stats app. I would like you to review and comment on the code added so far, since you have a lot more experience with views and urls.


backend/pong/constants.py
```diff
@@ -1,17 +1,26 @@
-# constants.py
+#Movement units, and angles
 PLAYER_MOVEMENT_UNIT = 1
+BALL_SPEED = 5 
+MAX_BOUNCE_ANGLE = 90  # Bounce angle in degrees
+MIN_BOUNCE_ANGLE = -90
```
merituulie, line 5: isn't better to use radians?

backend/pong/constants.py
```diff
@@ -1,17 +1,26 @@
...
+#Keys
 PLAYER_RIGHT_UP = 'o'
 PLAYER_RIGHT_DOWN = 'l'
 PLAYER_LEFT_UP = 'w'
-PLAYER_LEFT_DOWN = 's'
\ No newline at end of file
+PLAYER_LEFT_DOWN = 's'
+
+#Game Rules
+WINING_SCORE = 5
```
merituulie, line 26: typo, 'winning' with two 'n's

backend/pong/game_logic.py
```diff
@@ -30,9 +30,9 @@ def update_ball_position(self, pong_stat):
 
     def check_collisions(self, pong_stat):
         # Check for collisions with players
-        if pong_stat.ball.x <= pong_stat.player_left.x + PLAYER_WIDTH and pong_stat.player_left.y <= pong_stat.ball.y <= pong_stat.player_left.y + PLAYER_HEIGHT:
+        if pong_stat.ball.x <= pong_stat.player_left.x + PLAYER_WIDTH and pong_stat.player_left.y - PLAYER_HEIGHT / 2 <= pong_stat.ball.y <= pong_stat.player_left.y + PLAYER_HEIGHT / 2:
```
merituulie, line 33: smallest thing but still, as we know we always want to check PLAYER_HEIGHT / 2, we could just create a constant of that one too.

backend/stats/views.py
```diff
@@ -1,3 +1,28 @@
...
+from django.http import HttpResponse
+from rest_framework.response import Response
+from rest_framework import status
+# from django.utils.decorators import method_decorator
+# from User.serializers import UserOutputSerializer
 
-# Create your views here.
+# from shared_utilities.decorators import must_be_authenticated, \
+# 									must_be_body_user_id, \
+# 									valid_serializer_in_body
```
merituulie, line 13: nope for commented code!

backend/backend/urls.py
```diff
@@ -55,6 +57,9 @@
 	# Match Views
 	path('match/', MatchView.as_view()),
 	path('match/test/', LaunchTestMatchView.as_view()), # TEMPORARY
+
+	# Stats Views
+	path('stats/', StatsView.as_view()),
```
merituulie, line 62: So, what stats exactly does this return? Some fake data so far? I think we should separate different stats / data / graphs to their own routes. Like one route to return a graph / data formed from wins and another one from loses or something. But, you know better. For now I dont see any actual functionality here so its a little hard to say whats being added.

backend/stats/views.py
```diff
@@ -1,3 +1,28 @@
...
+
+class StatsView(APIView):
```
merituulie, line 15: Could and should be added later but will need the decorators. No need to add them now unless we actually start getting some data to build the stats from.

backend/pong/constants.py
dorianjagusch: Agreed. I'm working on the game rn because of some bugs and have changed that already

backend/pong/game_logic.py
dorianjagusch: Also taken care of 😂

backend/pong/constants.py
dorianjagusch: Taken care of in my next pr 😂

sakarisalmi (approved):
Nice start to the stats app!

backend/stats/views.py
azarSarikhani: removed!

backend/backend/urls.py
azarSarikhani: At this point yes fake data. I will add authentication and delivery of user's stats in the upcoming PRs. And in future PRs the url will also change into /stats/<int:user_id>

backend/stats/views.py
```diff
@@ -1,3 +1,55 @@
...
+        ]
+
+        # Create Plotly table figure
+		fig = go.Figure(data=[go.Table(header=dict(values=header),
+                                       cells=dict(values=cells))])
+
+        # Convert plot to HTML
+		plot_html = pio.to_html(fig, full_html=True)
+		full_html = f"""
+        <!DOCTYPE html>
```
merituulie, line 43: just a heads up for @dorianjagusch, we discussed that returning a html block is okay in terms of Restful API as long as we dont save the data / graph created here into the server - but a whole html page returned from here would break probably the single page application design, so removing the complete page as a response and in turn return the graph html would work. To not block further development we could merge this if everything else looks okay and as long as @azarSarikhani than continues with the graph development to fix the design breaking parts.

merituulie (approved):

PR #154, opened 2024-06-23
128 frontend pong

dorianjagusch:
# Finally!!!!!

Sorry for the delay. I could not be as productive as I wanted to be due to health bs. 
Enough excuses, however. Frontend pong is here. I apologise that it is a larger PR but many moving parts needed to fit together in order to get a MVP onto the main.
As of now, a user needs to login. After successful login, the url needs to be changed to /pong. 
As per Azzär's implementation, moving the right player starts the game.

### Additions:
- Use of Sakke's test API and token (PongService)
- Modularised pong game visualisation (2D)
- Websocket handling

### Todo:
- [ ] center and randomise ball direction
- [ ] standardize game object centers
- [ ] normalise field coordinates (client-side)
- [ ] integrate getting to pong into the page flow
- [ ] allow users to access pong api without login
- [ ] 3D pong
- [ ] close websocket on client side when navigating away in game


azarSarikhani (approved):
Thanks for cleaning up pong app files. Great work!

backend/Match/views.py
```diff
@@ -43,6 +43,6 @@ def get(self, request):
 		match, player_left_side, player_right_side = GameSetupManager.create_match_and_its_players(token)
 		if not all([match, player_left_side, player_right_side]):
 			return Response({'error': 'Something went wrong when creating the match and players'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
merituulie, line 45: Not really your issue but again, we should not CHOOSE to return a 500 :D

frontend/static/js/pages/profilePage.js
```diff
@@ -187,21 +202,23 @@ export default class extends AView {
...
+		);
 		const userDescription = profileDescription(userData.user.description);
 
-		const userStats = friendship === "friend"
-			? scrollContainer(
-				userData.stats,
-				profileStats, "row", "stat-list")
-			: null;
+		const userStats =
+			friendship === 'friend'
```
merituulie, line 213: Just a note for myself and us in the future, these equality checks should be done with the friendship enum instead of magic strings!

frontend/static/js/pages/pongPage.js
```diff
@@ -1,31 +1,29 @@
...
-		});
+		this.chatSocket = null;
+		this.pongService = new PongService();
 	}
 
 	async getHTML() {
+		let matchUrl = null;
+		try {
+			matchUrl = await this.pongService.getRequest();
+			this.chatSocket = new ChatSocket(matchUrl);
```
merituulie, line 20: Is it so that we stay in this page for the whole game, so we truly need to connect to a socket everytime one loads up this page? could one try to connect with another tab to the same socket etc?

frontend/static/js/pong/Player.js
```diff
@@ -0,0 +1,21 @@
...
+	}
+
+	createPlayer({players, game}) {
+		const playerGeometry = new THREE.BoxGeometry(players.width, players.height, 0);
+		const playerMaterial = new THREE.MeshBasicMaterial({color: 0xffffff});
+		const player = new THREE.Mesh(playerGeometry, playerMaterial);
+		if (this.isLeft) {
+			player.position.set(game.margin + players.width, game.height / 2, 0);
+		} else {
+			player.position.set(game.width - game.margin - players.width, game.height / 2, 0);
```
merituulie, line 16: do we get these measurements from the backend? or are we / should we make sure they match with the data from the backend or does the backend just send the middle coordinate of the player / ball?

dorianjagusch: the initial positions are just the center of the screen for the ball (that's in the next PR fixed), and for the players on the y-axis center. But even so, as the backend immediately after sends positions, those are corrected, if they don't already agree already

merituulie (approved):

PR #175, opened 2024-06-27
catch expired session

dorianjagusch:
Catch errors of expired session and added more exception classes to implement behaviour for cleaning data/redirecting upon 401 and 409 errors


merituulie (approved):

merituulie (approved):

hussojo (approved):
Looking good!

PR #179, opened 2024-06-27
178 tournament api part1 create models managers

sakarisalmi:
- created tournament app
- created Tournament and TournamentPlayer models
- created empty TournamentSetupManager and TournamentInProgressManager
- created tournamentMacros.py (used later) and tournamentState.py

not connected to tournament but matches:
- removed MatchManager as it wasn't used
- moved GameSetupManager from shared_utilities into Match/managers.py
  - renamed GameSetupManager to MatchSetupManager
- deleted shared_utilities/GameSetupManager.py


backend/Match/views.py
```diff
@@ -27,7 +27,7 @@ def get(self, request):
         if token.user_left_side.id != request.user.id:
             return Response({'error': 'You are not the host user in the token'}, status=status.HTTP_403_FORBIDDEN)
 
-        match, player_left, player_right = GameSetupManager.create_match_and_its_players(token)
+        match, player_left, player_right = MatchSetupManager.create_match_and_its_players(token)
 
         if not all([match, player_left, player_right]):
             return Response({'error': 'Something went wrong when creating the match and players'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
dorianjagusch, line 33: We need to return something other than 5xx in case something does in fact go wrong during the eval (like run out of space on the school computers or so)

backend/Match/views.py
```diff
@@ -40,7 +40,7 @@ class LaunchTestMatchView(APIView):
 	@method_decorator(must_be_authenticated)
 	def get(self, request):
 		token = MatchToken.objects.create_single_match_token(request.user, request.user)
-		match, player_left_side, player_right_side = GameSetupManager.create_match_and_its_players(token)
+		match, player_left_side, player_right_side = MatchSetupManager.create_match_and_its_players(token)
 		if not all([match, player_left_side, player_right_side]):
 			return Response({'error': 'Something went wrong when creating the match and players'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
dorianjagusch, line 45: Same here, not 5xx

backend/Tournament/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+	def abort_tournament(self):
+		if self.state == TournamentState.IN_PROGRESS.value:
+			self.state = TournamentState.ABORTED.value
+			self.end_ts = timezone.now()
+			self.save()
+
+	def __str__(self):
+		# The method to retrieve the human-readable representation of an IntegerChoices enumeration is get_FOO_display(), where FOO is the name of the field.
+		return f'Tournament {self.pk} - {self.get_state_display()}'
```
dorianjagusch, line 43: is pk from django? If we have controll would be nice to call it primary_key (if that's what it means)

backend/Tournament/models.py
```diff
@@ -0,0 +1,52 @@
+from django.db import models
+from django.utils import timezone
+
+from User.models import User
+from .tournamentState import TournamentState
+
+# Create your models here.
+class Tournament(models.Model):
+	host_user = models.ForeignKey(User, related_name='tournament_hosts', on_delete=models.CASCADE)
+	name = models.CharField(max_length=30, null=True, blank=True, default=None)
```
dorianjagusch, line 10: I see the 30 for char length a few times. Would be nice to have such magical values in a constants object or file, where they are easy to find and look up. Then I can catch them on the frontend already

backend/Tournament/exceptions.py
```diff
@@ -0,0 +1,13 @@
+from rest_framework.exceptions import APIException
+from rest_framework import status
+
+
+class TournamentSetupException(APIException):
```
dorianjagusch, line 5: Just food for thought: 
How do we want to handle when someone starts creating a tournament, lets the session expire in the middle of it and without refreshing sends further requests (e.g. completing player data).
Are they just kicked out and have to start over?

backend/Tournament/tournamentMacros.py
```diff
@@ -0,0 +1 @@
+TOURNAMENT_EXPIRERY_TIME_SECONDS = 60 * 20
```
dorianjagusch, line 1: expiry :P

dorianjagusch (commented):
With the 5xx errors removed, I'd be happy to approve

backend/Tournament/exceptions.py
sakarisalmi: when they'd try to send a new request with the token that they'd get an error informing them that the token has expired.

backend/Tournament/models.py
sakarisalmi: sure, but let's do it in some other pr

backend/Tournament/models.py
sakarisalmi: I changed it to self.id

backend/Match/views.py
sakarisalmi: changed

backend/Match/views.py
sakarisalmi: changed

backend/Tournament/tournamentMacros.py
sakarisalmi: oops :D

dorianjagusch (approved):

backend/Match/views.py
```diff
@@ -27,10 +27,10 @@ def get(self, request):
         if token.user_left_side.id != request.user.id:
             return Response({'error': 'You are not the host user in the token'}, status=status.HTTP_403_FORBIDDEN)
 
-        match, player_left, player_right = GameSetupManager.create_match_and_its_players(token)
+        match, player_left, player_right = MatchSetupManager.create_match_and_its_players(token)
 
         if not all([match, player_left, player_right]):
-            return Response({'error': 'Something went wrong when creating the match and players'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+            return Response({'error': 'Something went wrong when creating the match and players'}, status=status.HTTP_403_FORBIDDEN)
```
dorianjagusch, line 33: Technically this is not really it xD
Maybe 422?

dorianjagusch: Correction, maybe as in the Error 500 thread mentioned. We should find the right 5xx code. My mistake for misremembering the eval form

dorianjagusch: In school it would likely be 507, I'd wager (insufficient storage), but we should be confident of the cause in the different situations,  I'd say

sakarisalmi: we can talk about this more as a group later!

backend/Match/views.py
```diff
@@ -27,10 +27,10 @@ def get(self, request):
         if token.user_left_side.id != request.user.id:
             return Response({'error': 'You are not the host user in the token'}, status=status.HTTP_403_FORBIDDEN)
 
-        match, player_left, player_right = GameSetupManager.create_match_and_its_players(token)
+        match, player_left, player_right = MatchSetupManager.create_match_and_its_players(token)
```
merituulie, line 30: wait, you are throwing errors in the create_match_and ... function, but you are not catching them here, so we are probably anyway throwing a 500 if something goes wrnog there.

backend/Match/views.py
```diff
@@ -40,9 +40,9 @@ class LaunchTestMatchView(APIView):
 	@method_decorator(must_be_authenticated)
 	def get(self, request):
 		token = MatchToken.objects.create_single_match_token(request.user, request.user)
-		match, player_left_side, player_right_side = GameSetupManager.create_match_and_its_players(token)
+		match, player_left_side, player_right_side = MatchSetupManager.create_match_and_its_players(token)
```
merituulie, line 43: same here, we are not actually catching an error that might happen in the create_match... function are we

merituulie: if the problem is with postgres, we can return the 502 specifically but in any other case it should be some 400 but I dont see the forbidden being the correct one here.

backend/Tournament/exceptions.py
```diff
@@ -0,0 +1,13 @@
...
+    status_code = status.HTTP_403_FORBIDDEN
+    default_detail = 'There was an error creating the tournament or its players.'
+    default_code = 'tournament_creation_error'
+
+class TournamentInProgressException(APIException):
+    status_code = status.HTTP_403_FORBIDDEN
```
merituulie, line 11: Why are these using the forbidden status? that is an exception for when a resource that the requesting user is not allowed to access is used - which is a very specific case. Bad Request is usually the default one. please refer to the documentation about the status codes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses.

backend/Tournament/models.py
```diff
@@ -0,0 +1,52 @@
+from django.db import models
+from django.utils import timezone
+
+from User.models import User
+from .tournamentState import TournamentState
+
+# Create your models here.
+class Tournament(models.Model):
+	host_user = models.ForeignKey(User, related_name='tournament_hosts', on_delete=models.CASCADE)
```
merituulie, line 9: Does this mean with the cascade that we run into problems deleting the user to then dissolve the tournaments as well?

backend/Tournament/models.py
```diff
@@ -0,0 +1,52 @@
...
+	name = models.CharField(max_length=30, null=True, blank=True, default=None)
+	state = models.IntegerField(choices=TournamentState.choices, default=TournamentState.LOBBY)
+	next_match = models.IntegerField(default=0)
+	player_amount = models.PositiveIntegerField(null=False, blank=False)
+	winner = models.ForeignKey(User, related_name='tournament_winners', on_delete=models.CASCADE, null=True, blank=True, default=None)
+
+	insert_ts = models.DateTimeField(auto_now_add=True)
+	start_ts = models.DateTimeField(null=True, blank=True)
+	end_ts = models.DateTimeField(null=True, blank=True)
+	updated_ts = models.DateTimeField(auto_now=True)
```
merituulie, line 19: update_ts, as we talked about.

backend/Tournament/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+	expires_ts = models.DateTimeField(null=True, blank=True)
```
merituulie, line 21: Isnt this token stuff, a tournament doesnt really expire - or am I missing something?

backend/Tournament/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
+
+# Create your views here.
```
merituulie, line 3: Remove this please, we dont have views!

backend/Tournament/tournamentState.py
```diff
@@ -0,0 +1,7 @@
+from django.db import models
+
+class TournamentState(models.IntegerChoices):
+	LOBBY = 0, 'lobby'
+	IN_PROGRESS = 1, 'in_progress'
```
merituulie, line 5: we are using the '-' character rather than the underscore in enums.

merituulie:
Im still waiting for replies to other comments!

backend/Tournament/managers.py
```diff
@@ -0,0 +1,7 @@
+from .models import Tournament, TournamentPlayer
+
+class TournamentSetupManager:
```
merituulie, line 3: Eh we are again changing the logic, now we have different managers for different states of the tournament? Why not just one manager for the Tournament, no matter what the state of the tournament is.

backend/Match/models.py
```diff
@@ -19,8 +17,6 @@ class Match(models.Model):
 	created_at = models.DateTimeField(auto_now_add=True)
 	updated_at = models.DateTimeField(auto_now=True)
```
merituulie, line 18: updated -> update_ts here too actually!

backend/Tournament/models.py
merituulie: We also kind of break again the naming rule, why have insert_ts and then update_ts and suddenyl expire**S**_ts

merituulie (approved):
We can work on the status codes in the next PR / later when we get to finalizing the project.

merituulie (approved):

PR #182, opened 2024-06-27
107 standardise abstract classes

### LINKED ISSUE #114: Standardise async data handling
**Description**:
Currently we are using both .then.catch and async-await. They should be standardised to minimise context switching and make it overall easier to read the code.

Async-await makes for easier reading-flow, so I would go for that. Due to @merituulie restructuring of the services, using async-await would be my vote. While still introducing nesting when error-handling with try await, it allows for better reading flow, as long as we offload the work to the backend in structuring the data so we don't make multiple API calls to the backend to receive data for a page.

dorianjagusch:
Standardized abstract classes and error handling

The latter could be further improved by building on the error handler

closes #114


azarSarikhani (approved):

frontend/static/js/services/ArequestService.js
```diff
@@ -2,9 +2,9 @@ import constants from '../constants.js';
 import getCookie from '../utils/getCookie.js';
 import customErrors from '../exceptions/customErrors.js';
 
-class RequestService {
+class ArequestService {
```
merituulie, line 5: ARequest... so big R would maybe be a little bit more the basic PascalCase, no?

merituulie (approved):

PR #183, opened 2024-06-28
180 tournament api part2 setup tournament creation

sakarisalmi:
Building on PR 178.

Set up the creation of tournaments and the host tournament playesr.
setup the authenticating of guests and adding then to tournaments.
setup the removal the guest tournament players.
setup of launching of tournaments.
setup decorators that are utilised with tournaments.
Updated the OLD postman_collection.json that you can import to Postman and test the api. Sorry for not adding stuff to the new tests, but I'm quite tired as I didn't get any sleep last night.

You'll prob first want to check out the urls.py and then the Tournament/views.py to see what's going on exactly!


sakarisalmi:
refactored multiple tournament decorators into one.
some general small fixes as well

backend/Match/models.py
```diff
@@ -19,7 +12,8 @@ class Match(models.Model):
 	created_at = models.DateTimeField(auto_now_add=True)
 	updated_at = models.DateTimeField(auto_now=True)
 
-	objects = MatchManager()
+	tournament = models.ForeignKey(Tournament, related_name='matches', null=True, blank=True, default=None, on_delete=models.CASCADE)
+	tournament_match_id = models.PositiveIntegerField(default=0)
```
merituulie, line 16: Why two keys? Just one foreing key should be enoudh, and the cascading will probably cause na issue again with stats etc

backend/Tournament/managers.py
```diff
@@ -0,0 +1,174 @@
...
+from Player.models import Player
+
+import sys
+
+class TournamentSetupManager:
+    @staticmethod
+    def create_tournament_and_tournament_player_for_host(validated_data: TournamentCreationSerializer.validated_data):
+        try:
+            with transaction.atomic():
+                tournament = Tournament.objects.create(
```
merituulie, line 24: I think the creation of the tournament should be in the tournamentmanager and the creation of tournamentplayer in tournamentplayermanager. Then there would be some middle manager that will call both of these managers to create the two.

backend/Tournament/managers.py
```diff
@@ -0,0 +1,174 @@
...
+    @staticmethod
+    def setup_tournament_matches(tournament: Tournament) -> None:        
+        if tournament.state != TournamentState.LOBBY or tournament.next_match != 0:
+            raise TournamentSetupException("Tournament is not in a valid state to set up matchups.")
+        
+        tournament_players = list(tournament.players.all())
+        if len(tournament_players) != tournament.player_amount:
+            raise TournamentSetupException("The number of tournament_players does not match the expected player amount.")
+        
+        match_amount = tournament.player_amount - 1
```
merituulie, line 51: Is this dangerous, can the player amount be less than 1?

backend/Tournament/managers.py
```diff
@@ -0,0 +1,174 @@
...
+                    )
+
+        except Exception as e:
+            raise TournamentSetupException(f"An error occurred while setting up matchups: {e}")
+
+    @staticmethod
+    def start_tournament(tournament: Tournament) -> TournamentCreationSerializer:
+        player_amount = tournament.players.count()
+        if player_amount != tournament.player_amount:
+            raise TournamentSetupException(f"tournament must have {tournament.player_amount} players to start, but currently it has {player_amount} players.")
```
merituulie, line 87: Capital letter in the front please!

backend/Tournament/managers.py
```diff
@@ -0,0 +1,174 @@
...
+        TournamentSetupManager.setup_tournament_matches(tournament)
+        tournament.start_tournament()
+        tournament.save()
+
+
+class TournamentInProgressManager:
+    @staticmethod
+    def make_sure_tournament_has_not_expired(tournament: Tournament) -> None:
```
merituulie, line 95: abort_if_expired would be more descriptive name as this does something quite fatal!

backend/Match/models.py
merituulie: Ah I see... the column could be next_match etc, so just a foreing key to the next match.

backend/Tournament/views.py
```diff
@@ -0,0 +1,131 @@
...
+		tournament_creation_serializer = TournamentCreationSerializer(data=body_data)
+		if not tournament_creation_serializer.is_valid():
+			return Response({'error': 'invalid request for tournament creation'}, status=status.HTTP_400_BAD_REQUEST)
+		
+		try:
+			tournament = TournamentSetupManager.create_tournament_and_tournament_player_for_host(tournament_creation_serializer.validated_data)
+			host_tournament_player_serializer = TournamentPlayerSerializer(tournament.players.all().first())
+			return Response({
+				'tournament_id': tournament.id,
+				'tournament_player': host_tournament_player_serializer.data
```
merituulie, line 41: You should do like we do with the User and other things too, an Input and Output Serializers that expose these fields rather than creating it here by hand.

backend/Tournament/views.py
```diff
@@ -0,0 +1,131 @@
...
+			}, status=status.HTTP_201_CREATED) 
+		except Exception as e:
+			return Response({'error': str(e)},status=status.HTTP_403_FORBIDDEN)
```
merituulie, line 44: Forbidden is not a solve-all exception. What kind of exceptions can happen within this try block? catch those and depending on the exceptions return a proper status code.

backend/Tournament/views.py
```diff
@@ -0,0 +1,131 @@
...
+			except Exception as e:
+				return Response({'error': str(e)}, status=status.HTTP_403_FORBIDDEN)
+
+		elif tournament.state == TournamentState.IN_PROGRESS:
+			pass
+
+	@method_decorator(must_be_authenticated)
+	@method_decorator(check_that_valid_tournament_request([TournamentState.LOBBY, TournamentState.IN_PROGRESS]))
+	def delete(self, request, tournament_id):
+		tournament = Tournament.objects.filter(id=tournament_id).first()
```
merituulie, line 76: I dont think we need a delete route, or that the delete route should actually even delete anything but maybe just expire the tournament? But dont know for sure.

backend/Tournament/views.py
```diff
@@ -0,0 +1,131 @@
...
+		display_name = request.data.get('display_name', None)
+		if not username or not password:
+			return Response({"error": "Username and password are required"}, status=status.HTTP_400_BAD_REQUEST)
+		user = authenticate(username=username, password=password)
+		if user is None:
+			return Response({"message": "Invalid username or password"}, status=status.HTTP_401_UNAUTHORIZED)
+
+		tournament = Tournament.objects.get(id=tournament_id)
+		if tournament.players.count() >= tournament.player_amount:
+			return Response({"error": "Tournament already has maximum number of players"}, status=status.HTTP_403_FORBIDDEN)
```
merituulie, line 97: This is a bad request too

backend/backend/settings.py
```diff
@@ -148,7 +149,7 @@
 
 LANGUAGE_CODE = 'en-us'
 
-TIME_ZONE = 'UTC'
+TIME_ZONE = 'Europe/Helsinki'
```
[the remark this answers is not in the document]
merituulie: Probably does not matter here but just a heads up, for the db we usually want the UTC so no matter the timezone we are providing this service to, we can easily convert the UTC to local time.

backend/Tournament/managers.py
sakarisalmi: no because in the validation of the tournament creation serializer we check that the player_amount is either 4 or 8.

backend/Tournament/managers.py
sakarisalmi: sure

backend/Tournament/managers.py
sakarisalmi: sure

backend/Tournament/views.py
```diff
@@ -0,0 +1,131 @@
...
+											must_not_be_username, \
+											check_that_valid_tournament_request
+
+import sys
+
+# Create your views here.
+class TournamentListView(APIView):
+	@method_decorator(must_be_authenticated)
+	def post(self, request):
+		body_data = request.data.copy()
```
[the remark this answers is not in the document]
sakarisalmi: I use copy because you can't modify the request. so I just copy the the body and add the host user into the it before validating the serializer.

of course the frontend could just send the host user with the request, but we can get it in the backend regardless

backend/Tournament/views.py
sakarisalmi: the appropriate code would be 500, but we can't use that because of the eval. I'm using 403 in its stead, but if there is something more appropriate than that I'll use it

backend/Tournament/views.py
sakarisalmi: ok

backend/backend/settings.py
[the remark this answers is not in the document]
sakarisalmi: this was done because I was annoyed that the expires_ts wasn't showing the Finnish time, but I can revert it back to how it was earlier.

backend/Match/models.py
sakarisalmi: but how can it be a foreign key to next_match? we need to know in what order these matches are in the tournament.

backend/Tournament/managers.py
sakarisalmi: I've removed the tournament abortion from this function as it is done in the decorator that calls it. Now it just raises the exception.

backend/Match/models.py
merituulie: The same way as any other foreign key is the whole table?

merituulie: Like the same way we have foreign key to the tournament table we can have a foreign key to another row in the Match table.

backend/Tournament/managers.py
merituulie: Alright, then we just hope we dont forget this and try to save some possible invalid value in the db!

backend/Tournament/models.py
```diff
@@ -0,0 +1,52 @@
...
+from .tournamentState import TournamentState
+
+# Create your models here.
+class Tournament(models.Model):
+	host_user = models.ForeignKey(User, related_name='tournament_hosts', on_delete=models.CASCADE)
+	name = models.CharField(max_length=30, null=True, blank=True, default=None)
+	state = models.IntegerField(choices=TournamentState.choices, default=TournamentState.LOBBY)
+	next_match = models.IntegerField(default=0)
+	player_amount = models.PositiveIntegerField(null=False, blank=False)
+	winner = models.ForeignKey(User, related_name='tournament_winners', on_delete=models.CASCADE, null=True, blank=True, default=None)
```
merituulie, line 14: Again, not sure we should be using the CASCADE settings for these matches and such.

merituulie: https://sentry.io/answers/django-on-delete/

backend/Tournament/serializers.py
```diff
@@ -0,0 +1,128 @@
...
+
+class TournamentMatchSerializer(serializers.ModelSerializer):
+	first_tournament_player = serializers.SerializerMethodField()
+	second_tournament_player = serializers.SerializerMethodField()
+	winner = serializers.SerializerMethodField()
+	class Meta:
+		model = Match
+		fields = ['id', 'tournament_match_id', 'first_tournament_player', 'second_tournament_player', 'winner']
+
+	def get_first_tournament_player(self, match: Match) -> TournamentPlayerSerializer | None:
```
merituulie, line 29: first and second should be changes to left and right as its the way in other parts of this software!

backend/Tournament/serializers.py
```diff
@@ -0,0 +1,128 @@
...
+			raise serializers.ValidationError("This field may not be blank.")
+		if len(custom_name) > 30:
+			raise serializers.ValidationError("Ensure this field has no more than 30 characters.")
+		return custom_name
+	
+	def validate_tournament_player_amount(self, player_amount: int) -> None:
+		if player_amount not in [4, 8]:
+			raise serializers.ValidationError("Must have 4 or 8 players.")
+		
+	def validate(self, data):
```
merituulie, line 83: Nice!

backend/Tournament/serializers.py
```diff
@@ -0,0 +1,128 @@
...
+			'next_match',
+		]
+		
+	def get_next_match(self, tournament: Tournament) -> ReturnDict | None:
+		next_match = tournament.next_match
+		try:
+			next_match = Match.objects.get(tournament=tournament, tournament_match_id=next_match)
+			serializer = TournamentMatchSerializer(next_match)
+			# remove 'winner' from serializer
+			if 'winner' in serializer.fields:
```
merituulie, line 123: This seems a little hacky.. What and why are we doing here?

backend/Tournament/views.py
merituulie: 500 Is not appropriate way to ever answer from the Rest API - It means 'Oops something we didnt handle or think about is returned', as discussed multiple times before.

You should ask youserlf when coding: what kind of error can happen here? Depending on that you can then either catch different exceptions in there, or then use Bad request - but its also not fix-all solution, one should reflect why they are doing a try catch here and what they are trying to catch.

One should also think about when to use the try catch - its not something to just everywhere either.

merituulie: If you dont know what can happen, you can just remove the try catch block and see what happen, we'll catch the possible error when / if such a thing can happen here.

backend/Tournament/views.py
```diff
@@ -0,0 +1,131 @@
...
+		try:
+			tournament_player = TournamentPlayer.objects.create(
+				tournament=tournament,
+				user=user,
+				display_name=display_name,
+			)
+			tournament_player_serializer = TournamentPlayerSerializer(tournament_player)
+			return Response(tournament_player_serializer.data, status=status.HTTP_201_CREATED)
+		except Exception as e:
+			return Response({'error': str(e)}, status=status.HTTP_403_FORBIDDEN)
```
merituulie, line 108: And please check all of the other errors too, forbidden is very rearly the correct status code.

backend/backend/settings.py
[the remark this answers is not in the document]
merituulie: will it be the same in the UI and the database? Thats the most important thing. If the database clock and frontend timezones are different and we dont handle it, it can cause a lot of issues with these expiring things!

backend/shared_utilities/decorators.py
```diff
@@ -100,3 +112,40 @@ def wrapper(*args, **kwargs):
...
+
+			if request.user != tournament.host_user:
+				return HttpResponseForbidden("You are not authorized to modify this resource.")
+
+			if tournament.state not in tournament_states:
+				return HttpResponseForbidden(f"Cannot modify this resource because tournament state is not in '{tournament_states}'.")
+			
+			try:
+				TournamentInProgressManager.make_sure_tournament_has_not_expired(tournament)
+				TournamentInProgressManager.make_sure_users_in_active_tournament_are_still_active(tournament)
```
dorianjagusch, line 143: see above

backend/shared_utilities/decorators.py
```diff
@@ -100,3 +112,40 @@ def wrapper(*args, **kwargs):
...
+				return HttpResponseNotFound(f"Tournament {tournament_id} does not exist.")
+
+			if request.user != tournament.host_user:
+				return HttpResponseForbidden("You are not authorized to modify this resource.")
+
+			if tournament.state not in tournament_states:
+				return HttpResponseForbidden(f"Cannot modify this resource because tournament state is not in '{tournament_states}'.")
+			
+			try:
+				TournamentInProgressManager.make_sure_tournament_has_not_expired(tournament)
```
dorianjagusch, line 142: away with "make_sure". is_tournament_active/ check_tournament_active/

dorianjagusch (changes requested):
The naming and as Meri pointed out, error handling are a bit all over the place. 
To test the individual pieces during development, it might help to take either the must be authenticated or tournament tokens out, just to simplify the debugging. git stash or patch files, as discussed some weeks ago on discord, can be your best friends with this.

backend/shared_utilities/decorators.py
merituulie: A general convention is usually indeed:
- if its a if statement based boolean function, starts with 'is' or 'are' and returns if true or not
- if a try catch, starts with 'try'
- if its a if statement based function that does something in a true / false case -> 'delete_if_' etc..

In general for boolean functions there is a rule in Clean Code by Uncle Bob: _"Avoid negative name for boolean function"_, its weird to assume a true result from a function called 'does_NOT_fulfil_the_if'. Refer to for example: https://levelup.gitconnected.com/knot-of-nots-avoiding-negative-names-for-boolean-methods-641896a94a42

backend/Match/models.py
```diff
@@ -10,10 +10,9 @@ class Match(models.Model):
 	start_time = models.DateTimeField(null=True, blank=True)
 	end_time = models.DateTimeField(null=True, blank=True)
 	created_at = models.DateTimeField(auto_now_add=True)
-	updated_at = models.DateTimeField(auto_now=True)
+	update_at = models.DateTimeField(auto_now=True)
```
dorianjagusch, line 13: Just for my own understanding: Why is created at in past tense and update in present tense?

dorianjagusch (approved):

merituulie (approved):

PR #186, opened 2024-07-07
181 delete user from backend

hussojo:
Added backend logic to deleting user and couple postman test.


backend/User/views.py
```diff
@@ -79,7 +80,17 @@ def delete(self, request, user_id):
 			return Response(status=status.HTTP_404_NOT_FOUND)
```
dorianjagusch, line 80: I know it's not part of this PR, but shouldn't we first check that the request is well-formed and only if it is log the user out/ send a 404?

backend/User/views.py
```diff
@@ -79,7 +80,17 @@ def delete(self, request, user_id):
...
 		except:
 			return Response(status=status.HTTP_400_BAD_REQUEST)
-		user.delete()
+		user.username = get_random_string(length=30)
```
dorianjagusch, line 83: I don't know what was discussed yesterday but we might want do to something else than a random string here like "deleted user" or something, I would say (unless there's a good reason to randomise it)

dorianjagusch (approved):

dorianjagusch:
Once my PR comes through you can "easily" add a confirmation popup to our page with authentication. Both modals are basically complete for that. We might need to discuss about the services and how to connect them in a flow, cause that can be slightly counter intuitive (decoupling wohooo)

dorianjagusch (approved):

backend/User/views.py
```diff
@@ -79,7 +80,17 @@ def delete(self, request, user_id):
...
+		user.set_password(get_random_string(length=30))
+		user.is_active = False
+		user.is_staff = False
+		user.is_superuser = False
+		user.insertTS = None
+		user.last_login = None
+		user.is_online = False
+		# profile_picture = None
```
merituulie, line 91: the profile_picture comment can be taken out, itll be in its own table!

merituulie (approved):

backend/User/views.py
hussojo: As far as I understand the methods used are already making sure user is logged in and also the owner. So the 404 should never even happen in this case 🤷‍♀️

dorianjagusch (approved):

PR #187, opened 2024-07-08
51 add ability to use avatar to the user

merituulie:
Okay, finally the basic functionality to POST a new profile picture and GET it from the backend. Tried to clean up a little bit but its... hacky to say the least.

The functionality works currently just for the sidebar, so next up would be to use this functionality every where we need the profilepicture in. Here's a snippet of the functionality:

https://github.com/dorianjagusch/ft_transcendence/assets/44494010/2ff729dd-53a9-4d52-b78e-9bc191fa10ce


backend/User/managers.py
merituulie, line 1: Oh, I removed this is_staff stuff as we dont really use it anywhere. Hope thats okay for @sakarisalmi !

backend/User/views.py
```diff
@@ -95,7 +138,7 @@ def post(self, request):
 			login(request, user)
 			# set additional session data if necessary
 			request.session['is_authenticated'] = True
```
merituulie, line 140: In order to use the localStorage in the UI to get for example logged in users id in order to make new requests for the user by its id, we return the user from the login!

frontend/static/js/components/profilePicture.js
```diff
@@ -0,0 +1,46 @@
...
+				return '';
+			}
+			else {
+				throw new Error(`Error: ${response.status}`);
+			}
+		}
+
+		return response.json();
+	})
+	.then(responseBody => {
```
merituulie, line 30: Mmmeh, we are now using a promise here since the async would bleed the try catching into the sidebar and just felt wrong. Hopefully its okay to do the promise here for the sake of I just dont want to go back and fight this :D

azarSarikhani (approved):

backend/User/validators.py
```diff
@@ -0,0 +1,17 @@
+from django.core.exceptions import ValidationError
+import os
+
+def validate_image(file):
+	valid_mime_types = ['image/jpeg', 'image/png', 'image/gif']
```
dorianjagusch, line 5: gifs? really? xD

backend/User/views.py
dorianjagusch: This is just for the logged in user? because you return the friend already, right?

on a separate note. what url should I display to the user, when they are on a friends page? Do we want them to even navigate to their friends via the url bar? I will make an issue for this

backend/User/views.py
```diff
@@ -95,7 +145,7 @@ def post(self, request):
...
-			return Response({"message": "User login successful"}, status=status.HTTP_202_ACCEPTED)
+			return Response(UserOutputSerializer(user).data, status=status.HTTP_202_ACCEPTED)
```
dorianjagusch, line 148: I don't know how django does it, but couldn't we have a message nonetheless to give context about what what successful?

backend/backend/settings.py
```diff
@@ -174,6 +179,7 @@
 
 CSRF_TRUSTED_ORIGINS = [
 		'http://localhost:80',
+		'http://localhost:8080',
```
dorianjagusch, line 182: Is the backend sending requests to itself?

frontend/static/css/styles.css
```diff
@@ -983,10 +983,10 @@ button:is(:focus, :hover) {
 .profile.dashboard {
 	grid-template-areas:
 		'title title title title title title title'
-		'profile-img profile-img placements placements placements placements placements'
-		'about about about stats stats stats stats'
-		'history history history graph-1 graph-1 graph-1 graph-1'
-		'history history history graph-2 graph-2 graph-2 graph-2';
+		'profile-img profile-img placements placements placements summary summary'
+		'about about about stats-1 stats-1 stats-2 stats-2'
```
dorianjagusch, line 987: Oh I need to remove the second game from friends still. Another issue

frontend/static/js/components/profilePicture.js
dorianjagusch: It's fine for now, I would say. I am not 100% I understand what you mean with bleeding into the side bar, but either way we'll figure it out once the proper loading of the image everywhere is in place to have a unified handling of that

dorianjagusch: I have started working on a unified error handling thing for all kinds of errors coming in (hence also the above comment regarding the error messages in responses) and in an optimal world we should be able to handle this elegantly

dorianjagusch (approved):
Looks good to me. I am slightly worried about conflicts or old changes coming back from the abyss, since there were some changes in files that don't exist anymore in their old locations, but we''ll have to just see when we cross that bridge :D

backend/User/views.py
merituulie: Yup, this is for the logged in user, friends will get their pictures on a different route.

frontend/static/js/components/profilePicture.js
merituulie: Yeah, honestly Im kind of embarrased I could not get it to work properly and that is mostly just me not being sure what would work but Im hoping we can go forward with this being its just here in this function and fix it later together or smt - or then just let it be and pass :D I trust you though so if you're cool with this, nice. If not, lets check it out futher. I am making changes in the next PR too so maybe its premature to change at this point.

backend/User/validators.py
merituulie: :D I know right, didnt test it yet but on the next PR will :D

backend/User/views.py
```diff
@@ -82,6 +87,51 @@ def delete(self, request, user_id):
...
+			if not profile_picture:
+				return Response(status=status.HTTP_404_NOT_FOUND)
+
+			# Get the actual file path from the ImageFieldFile instance
+			image_path = profile_picture.picture.path
+			with open(image_path, "rb") as image_file:
+				image_data = image_file.read()
+				encoded_image = base64.b64encode(image_data).decode('utf-8')
+				return Response({'image': encoded_image}, status=status.HTTP_200_OK)
+		except User.DoesNotExist:
```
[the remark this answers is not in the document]
merituulie: Ah yeah. This is actually a little bit weird of a route as we handle in the UI basically the fact that something is not found. Basically this should never get here without a user that is actually logged in an existing so I was being lazy in this regard. What we do in the next PR is actually return an empty string and OK because in 90% of the cases it means the user did not load their own avatar -> we use the default one and no need to return an actual error :D But I am open to concerns! Jsut felt like returning error messages might also give false positives kind of.

backend/backend/settings.py
merituulie: Shit, no, I got the auth issues with the localstorage that made me try all kinds of things and forgot 🗡️

backend/User/validators.py
```diff
@@ -0,0 +1,17 @@
...
+
+	ext = os.path.splitext(file.name)[1].lower()
+	if ext not in valid_file_extensions:
+		raise ValidationError('Unsupported file extension. Allowed extensions are: .jpg, .jpeg, .png, .gif.')
+
+	if file.size > max_file_size:
+		raise ValidationError('File size exceeds the maximum limit of 2MB.')
+
+	if file.content_type not in valid_mime_types:
+		raise ValidationError('Unsupported file type. Allowed types are: image/jpeg, image/png, image/gif.')
```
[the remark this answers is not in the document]
merituulie: nvm, removed them from the message :D

frontend/static/js/components/profilePicture.js
dorianjagusch: Guuuurl, you should have a look at the fucked up shit I did in my PR just trying to keep the different modals independent from their surroundings and to keep them as modular as I could xD

dorianjagusch (approved):

PR #191, opened 2024-07-10
158 3D pong

dorianjagusch:
Boing boing boing

Pong in 3D is here!

## Changes
- Added dimension switch by pressing space bar
- refactored PongGame
- expanded player and ball
- added lights and plane

## Caveats
-  Ball and Player should be one class or inherit from one common class :see_no_evil:  FIXED
- There are some glitches still on the backend


.gitignore
```diff
@@ -156,4 +156,6 @@ GitHub.sublime-settings
 *.patch
 
 #Ignore personal notes
-notes.txt
\ No newline at end of file
+notes.txt
+**/node_modules/
+**/*package*.json
```
merituulie, line 161: since we are ignoring uploading these, is loading the needed packages added to the makefile?

frontend/static/js/pong/Light.js
```diff
@@ -0,0 +1,42 @@
...
+		if (!(this.light instanceof THREE.SpotLight)) {
+			throw new TypeError('Light must be a SpotLight to follow an object.');
+		}
+		this.light.target = target;
+		this.light.target.updateMatrixWorld();
+	}
+
+	moveLightCircular() {
+		if (this.angleRad === null || this.speed === null) {
+			throw new NotImplentedError('angleRad and speed must be set to move light circularly.');
```
merituulie, line 36: a small thing but NotImplementedException is a quite special exception in specifically usually indicates a future feature for a feature! Like for example now that we inherit the ARequestService we are lazy and use that to block some function calls through services that dont handle a specific request / we wont expose a specific api call with specific service in this project -> check out GET from FriendService!

More appropriate might be something like ArgumentException (assuming we wont really ever get here and wont show the error to the user) - tells the coder that you are trying to use this function with faulty arguments.

frontend/static/js/pong/Player.js
```diff
@@ -1,22 +1,55 @@
...
+	switchMaterial() {
+		if (this.player.material === this.materials.default) {
+			this.player.material = this.materials.alternative;
+			this.player.receiveShadow = true;
+			this.player.castShadow = true;
+		} else {
+			this.player.material = this.materials.default;
+			this.player.receiveShadow = false;
+			this.player.castShadow = false;
+		}
```
merituulie, line 42: Might make sense to just pass the material and the boolean value here and set it

.gitignore
dorianjagusch: That's just for now to not clutter up the main with unused stuff

frontend/static/js/pong/Light.js
dorianjagusch: I could write our own error for that (JS doesn't come with one out of the box). I would do it with the more centralised error handling in another PR

azarSarikhani (approved):

azarSarikhani (approved):

Issue #185, opened 2024-07-04
Make sure the side bar is updated at every point


merituulie:

dorianjagusch:
We could use this issue to go further and make a request to the backend from the router to retrieve up-to-date user information (username, loginstatus, profilePicture, userId)

PR #192, opened 2024-07-10
188 extend avatar usage in the front end

merituulie:
Extended the avatar to update the dashboard picture and the navigation bar picture. Needs a page refresh though to work.

Also added sessionStorageManagement to empty out localStorage to fix the issue with closing the browser and opening it again - doenst work in all cases yet but with refreshing the page we will get to fix that too. Reloading a page persists the logged in user though as it should.

### DIFF
--- a/frontend/static/js/views/logoutView.js
+++ b/frontend/static/js/views/logoutView.js
@@ -12,6 +12,8 @@ export default class extends AView {
 		try {
 			 await logoutService.postRequest();
 			localStorage.clear();
+			localStorage.setItem('isLoggedIn', false);
+			sessionStorage.setItem('isLoggedInSession', false);
 			document.cookie.split(';').forEach((c) => {
 				document.cookie = c
 					.replace(/^ +/, '')
--- a/backend/Friends/views.py
+++ b/backend/Friends/views.py
@@ -18,6 +18,9 @@ class FriendsListView(APIView):
 			return Response({"message": "Expected to get parameter 'friendship_status'"}, status=status.HTTP_400_BAD_REQUEST)
 		user_id = request.user.id
 		friends = Friend.objects.get_user_friends(user_id, friendship_status)
+
+		if not friends:
+			friends = []
 		serializer = UserOutputSerializer(friends, many=True)
 		return Response(serializer.data, status=status.HTTP_200_OK)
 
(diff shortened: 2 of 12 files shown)


azarSarikhani (approved):

frontend/static/js/views/logoutView.js
```diff
@@ -12,6 +12,8 @@ export default class extends AView {
 		try {
 			 await logoutService.postRequest();
 			localStorage.clear();
+			localStorage.setItem('isLoggedIn', false);
```
dorianjagusch, line 15: I am confused. What does the sessionStorage do that the localStorage doesn't do? We should also make a clear decision about what we want the behaviour to be when people close the tab/window and revisit the site. Some things will obviously change once all pages properly make the requests they need to (I think all but the play page)

dorianjagusch (approved):

frontend/static/js/views/logoutView.js
merituulie: I know its a bit wonky in logic still, I wasnt too sure so went in with some hazy shit. But now I'm thinking we should actually use the sessionstorage rather than localStorage for stuff, here are the lifetime differences:
- data stored in localStorage persists even after the browser is closed and reopened. It only gets cleared if the user explicitly deletes it through browser settings or via JavaScript.
- data stored in sessionStorage is cleared when the page session ends. This usually happens when the tab or window is closed. However, if the user reloads the page or navigates to a different page within the same tab, the data remains available.

merituulie: I think the login and information should persist when opening other tabs and going back and forth with the browser open. However, when we completely close the browser, I think it should clear the info. But that is also from the point of view of debugging while developing made easier - for me at least. I didnt dig too deep into this as clearing the data using sessionStorage seemed to work well for the case I needed it for, so definitely am open to more views on it!

dorianjagusch: I think so to, especially in terms of UX, since that is at least for me the expected behaviour. So localStorage it is. When the loading of the frontend works as I picture it (for that the API needs to work for the different views), basically every page makes requests and the frontend would catch the 401 and clear the storage and "log" the user out. Alternatively, and also to handle the play and match views, which only send requests after a modal has been filled out and submitted, we could just send a request to check that we're still logged in every time we go through the router. There we could also retrieve the profile picture every time. We could do this in #185.

PR #194, opened 2024-07-10
Added proper websocket closing when user navigates away

dorianjagusch:
Another day another win!

Now navigating away from the pong page, be it by browser navigation, url or the navbar, the socket is being closed and the evenhandlers removed to prevent them from doing the shenanigans they did. 

## Changes
- Added an event listeners on the navbar, indirectly on the browser buttons and lastly   a mutation observer that continuously looks for changes in the DOM. When detected whether it is the element with id "pong" (given that there is only one pong page and ids are unique anyway), it finds, disables and overrides the websocket :D

## Caveats
The event appending on the navbar could be improved by taking advantage of event delegation

### DIFF
--- a/frontend/static/js/views/pongView.js
+++ b/frontend/static/js/views/pongView.js
@@ -11,6 +11,29 @@ export default class extends AView {
 		this.setTitle('Pong');
 		this.chatSocket = null;
 		this.pongService = new PongService();
+		this.attachEventListeners = this.attachEventListeners.bind(this);
+	}
+
+	attachEventListeners() {
+		document.querySelectorAll('.nav-link').forEach((link) => {
+			link.addEventListener('click', this.chatSocket.handleClose);
+		});
+
+		window.addEventListener('beforeunload', this.chatSocket.handleClose);
+
+		const observer = new MutationObserver((mutationsList, observer) => {
+			for (let mutation of mutationsList) {
+				if (mutation.removedNodes) {
+					mutation.removedNodes.forEach((node) => {
+						if (node.id = "pong") {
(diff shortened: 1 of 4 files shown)


frontend/static/js/views/pongView.js
```diff
@@ -2,15 +2,36 @@ import ChatSocket from '../pong/ChatSocket.js';
...
+	}
+
+	attachEventListeners() {
+		document.querySelectorAll('.nav-link').forEach((link) => {
+			link.addEventListener('click', this.chatSocket.handleClose);
+		});
+
+		window.addEventListener('beforeunload', this.chatSocket.handleClose);
+
+		const observer = new MutationObserver((mutationsList, observer) => {
```
merituulie, line 22: again the smallest thing ever but quite the nesting you got going on here :P will reverting the ifs make this a little bit more readable?

merituulie (approved):

azarSarikhani:
This does not close the socket on the backend though, does it? could we send a "q" for instance to the back and add to the back if key received is "q" close the socket or disconnect?

dorianjagusch:
Indeed it does not close it on the backend. 

I would rather suggest ESC, which could also just quit the game in general. "Q" is dangerously close to "W" :sweat_smile:

hussojo (approved):
luv it

frontend/static/js/views/pongView.js
dorianjagusch: Just noticed this comment now. I marked it as a todo for some clean up pr, I plan for the end of the week

PR #197, opened 2024-07-14
Player selection modal

dorianjagusch:
I have been fighting with spliting the PR, but I had not been diligent enough with small commits in between either... 
If this PR is way to big and you don't want to review it like this. I can do the nooby stuff and copy the affected files individually in to make a cleaner history, but I thought I'd try my luck first :P

So the important thing to understand here is that the main buttons of a modal (continue/back) are controlled by the whoever calls the modal. For most it's the page, but for the authentication modal, it can be the playerSelectionModal.

Data from one modal to the caller is done by passing the function into the constructor of the modal. The modal is thus just responsible to get the data and the parent chooses what to do with it itself. 

The Forms as in the Match/Views PR, are passed as well and are independent of any services. The modal, with the service attached to it decides what  data is used to call it's service and passes the data to the parent, which decided the service and form to begin with. 

The code is a little untidy, mostly due to html stuff. I tried and kept the page/modal specifications in their own methods, but they do pollute the class a little. If anyone has an idea on how to get rid of that, while letting the caller decide the look and tags, lmk. 

As a little help to understand the architecture, a little sketch:
https://www.figma.com/board/qSdKtcEwoAxz9T2vQEB0ex/Untitled?node-id=0-1&t=dzBtVI3ok8LDv0Vc-0

## Changes: 
- Created tournament naming and player number modal
- Created modal for summary of tournament set up
- Created modal for player selection, including calling the authentication modals
- moved modal notification into the Abstract class

## Caveats:
Needs to properly handle the data coming from the tournament and match APIs once they are complete


dorianjagusch:
https://github.com/user-attachments/assets/120babbe-538e-4694-bc09-b0f4d412a808

And if it works, there's also a video (had to use some sketchy converter. I can try again properly tomorrow again, if this turns out badly)

frontend/static/js/components/dialogs/tournamentModals/tournamentModal.js
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	async createTournament(formData) {
+		const tournament = {
+			custom_name: formData.tournamentName,
+			id: 1,
+			player_amount: formData.numberOfPlayers,
+			player_ids: [1, 5, 3, 2, 6, 4, 7]
+		};
+		// return await this.service.postRequest(tournament);
```
merituulie, line 28: who dis

merituulie: ah mock shit

frontend/static/js/components/formComponents/selectPlayersForm.js
```diff
@@ -0,0 +1,64 @@
...
+		const playerList = document.createElement('section');
+		playerList.classList.add('player-list', `player-${this.numberOfPlayers}`);
+		form.appendChild(playerList);
+
+		for (let i = 0; i < this.numberOfPlayers; i++) {
+			if (i == 0) {
+				const playerCard = PlayerCard(
+					localStorage.getItem('username'),
+					// localStorage.getItem('userImage')
+					'../static/assets/img/default-user.png',
```
merituulie, line 26: if my PR happens to go thourgh, maybe would make sense to use the getFriendProfilePicture here, will give out the img attirubte to be appended to the card!

merituulie (approved):

frontend/static/js/components/formComponents/selectPlayersForm.js
dorianjagusch: I'll wait then 😎

sakarisalmi (approved):
LGTM

PR #200, opened 2024-07-16
193 extend getting friend avatars

merituulie:
Had to do some duplicate code but too tired to try to make it better this iteration :P scrollers now use the given element.img as a whole and dont create the img element separately.


frontend/static/js/router.js
```diff
@@ -56,4 +56,10 @@ const router = async () => {
 	view.getHTML();
 };
 
+window.addEventListener('load', () => {
+	if (!sessionStorage.getItem('isLoggedInSession')) {
```
dorianjagusch, line 60: instead of this we could have the router make a request to the backend for the supposedly logged in user and if it get a 401, we're logged out (that happens already). I will also implement that the current page the user tried to access will be loaded after logging in. This is more of a comment on what's to come than anything for you to change :P

dorianjagusch (approved):

dorianjagusch (approved):

sakarisalmi (approved):
nice!

PR #206, opened 2024-07-17
201 add user search by username fragments

dorianjagusch:
Search of users by (partial) username works now.

It is probably not the cleanest solution. There is a bit too much logic in the Output serializer for my taste. If it's indeed to much for people and if the interdependence of the apps is too much for people, I am happy to change this with the help of our django experts. 

The current character limit until the search is started is 3 chars and only on the frontend!

## Changes:
- Added authentication barrier conditional branch to GET user/ endpoint 
- remodelled UserFriendOutputSerializer to work with querysets as well as singular objects
- Integrated user service with frontend to make use of the optional queryparameter

## Caveat:
- User image is currently not being retrieved until #193 is through

closes #201

PR #214, opened 2024-07-18
resize button images

dorianjagusch:
Button images are resized to fit better into the buttons (easier to test once #206 is through and merged into this)


sakarisalmi (approved):
Nice

PR #220, opened 2024-07-19
208 add ai as the right player to backend

merituulie:
Added AI Opponent. Can be tested with chaning the ai_opponent boolean in consumers.py to true! There is a lot of values that one can try out to find a good compromise with being enough of an opponent and smooth moving on the grapics but, its good enough compared to the fact it can only evaluate the balls future position once a second.


sakarisalmi (approved):
Great stuff!

dorianjagusch (approved):
Noice

Issue #195, opened 2024-07-11
Fix backend error codes


sakarisalmi:
Check and fix backend error codes especially when it comes to the tournament api.

dorianjagusch:
Is this done?

PR #221, opened 2024-07-19
184 tournament api part3 handle matches set up tournamentserializer

sakarisalmi:
I don't know if my previous comment can be seen when I reopen this pr, but just in case I'll write the same stuff here again.

This is the last tournament API PR. There'll probably be some later changes here and there, but with this PR the infrastructure for playing the tournaments from start to finish is set up!

- Adds the ability to play tournament matches
- refactored TournamentMatchSerializer so that it has both the tournament_match_id (i.e. the order in which the matches are played in the tournament) and the match_id
- removed 'tournamentplayers' and 'matches' from TournamentInProgressSerializer; next_match is just the next match to be played (the tournament_match_id, NOT match_id)
- created TournamentMatchListView that returns serializer of all the tournament matches of a specific tournament
- created TournamentMatchDetailView that returns serializer of a specific tournament match
  - will return match url so that the match can be played IF
    - the request is sent by the host
    - the tournament is in progress
    - the tournament_match_id is the next match in the tournament
- created TournamentPlayersListView that returns serializer of all the players of a specific tournament
- refactored pong/consumers.py so that it updates tournament if played match is a tournament match

other small changes:
- MatchTokenManager is not linked directly to MatchToken model; refactored token views to take this into account
- MatchSerializer now has a 'tournament' field
- MatchSerializer's 'state' is now the string value of the state (for example: 'state': 'lobby' or 'state': 'in-progress'); removed state_display


backend/Tournament/managers.py
```diff
@@ -99,6 +99,26 @@ def raise_error_if_inactive_user_in_tournament(tournament: Tournament) -> None:
         inactive_users = tournament_players.filter(user__is_active=False)
         if inactive_users.exists():
             raise TournamentInProgressException(f"An user in the unfinished tournament has deleted their account")
```
dorianjagusch, line 101: A user*

backend/Tournament/serializers.py
```diff
@@ -19,12 +19,20 @@ class Meta:
...
+	tournament_match_id = serializers.SerializerMethodField()
+	match_id = serializers.IntegerField(source='id', read_only=True)
+	state = serializers.CharField(source='get_state_display', read_only=True)
 	tournament_player_left = serializers.SerializerMethodField()
 	tournament_player_right = serializers.SerializerMethodField()
 	winner = serializers.SerializerMethodField()
 	class Meta:
 		model = Match
-		fields = ['id', 'tournament_player_left', 'tournament_player_right', 'winner']
+		fields = ['tournament_match_id', 'match_id', 'state', 'tournament_player_left', 'tournament_player_right', 'winner']
```
dorianjagusch, line 30: for the individual match_id, id is enough, I would say. Although, can't hurt to be explicit, I suppose

backend/Tournament/managers.py
sakarisalmi: shouldn't it be 'an' though? the user starts with a vowel.

backend/Tournament/views.py
```diff
@@ -51,34 +54,34 @@ def get(self, request, tournament_id):
 		return Response(serializer.data, status=status.HTTP_200_OK)
 
 	@method_decorator(must_be_authenticated)
-	@method_decorator(validate_tournament_request([TournamentState.LOBBY, TournamentState.IN_PROGRESS]))
+	@method_decorator(validate_tournament_request([TournamentState.LOBBY]))
 	def post(self, request, tournament_id):
-		'''If tournament is in lobby, start tournament.
-
-		If tournament is in progress, return next match url.
+		'''starts the tournament.
```
dorianjagusch, line 59: Either use docstrings to describe the function properly, or leave it out. We have established the logic already (and if it follows REST, it's almost self-explanatory). If you want to leave me hints, instead of hidden in the code, a sequence diagramme or proper documentation would be appreciated (postman, can also help with that). 

TL;DR no docstring needed here :D

backend/Tournament/views.py
```diff
@@ -12,6 +12,9 @@
 							TournamentInProgressException
 from .tournamentState import TournamentState
 from User.models import User
+from Tokens.models import MatchToken
+from Tokens.managers import MatchTokenManager
+from Match.matchState import MatchState
 from shared_utilities.decorators import must_be_authenticated, \
 											must_not_be_username, \
 											validate_tournament_request
```
dorianjagusch, line 20: I can comment on line 22 nd 24, but those should be deleted

backend/Tournament/views.py
```diff
@@ -123,4 +126,50 @@ def delete(self, request, tournament_id, tournamentplayer_id):
...
+		tournament = Tournament.objects.filter(id=tournament_id).first()
+		if not tournament:
+			return Response(f"Tournament {tournament_id} not found.", status=status.HTTP_404_NOT_FOUND)
+		tournament_matches = tournament.matches.all().order_by('id')
+		tournament_match_serializers = TournamentSerializers.match(tournament_matches, many=True)
+		return Response(tournament_match_serializers.data, status=status.HTTP_200_OK)
+
+class TournamentMatchDetailView(APIView):
+	def get(self, request, tournament_id, tournament_match_id):
+		'''get details of a single tournament match.
```
dorianjagusch, line 142: See above

backend/Tournament/views.py
```diff
@@ -123,4 +126,50 @@ def delete(self, request, tournament_id, tournamentplayer_id):
...
+		'''
+		tournament = Tournament.objects.get(id=tournament_id)
+		if not tournament:
+			return Response(f"Tournament {tournament_id} not found", status=status.HTTP_404_NOT_FOUND)
+		tournament_matches = tournament.matches.all().order_by('id')
+		if tournament_match_id >= tournament_matches.count():
+			return Response(f"Tournament {tournament_id} does not have {tournament_match_id} match", status=status.HTTP_404_NOT_FOUND)
+		
+		# check if user is host, tournament is in progress, and the tournament_match_id is the next_match, if so, start match
+		if request.user == tournament.host_user \
```
dorianjagusch, line 153: A long if statement like that could be pulled out into a function with a descriptive name. Makes the comment obsolete and the code easier to read

backend/Match/managers.py
```diff
@@ -12,7 +12,6 @@ class MatchSetupManager:
 	def create_match_and_its_players(match_token: MatchToken) -> Match | None:
 		if not isinstance(match_token, MatchToken):
 			raise TypeError((f'match_token must be an MatchToken, not {type(match_token).__name__}'))
```
dorianjagusch, line 14: superfluous since you the type is already being checked

dorianjagusch, line 14: the if statement is superfluous due to type checking

backend/Tournament/views.py
```diff
@@ -123,4 +126,50 @@ def delete(self, request, tournament_id, tournamentplayer_id):
...
+			and tournament.state == TournamentState.IN_PROGRESS \
+			and tournament_matches.filter(state=TournamentState.FINISHED).count() == tournament_match_id:
+
+			# check that the tournament is still valid
```
dorianjagusch, line 157: Unnecessary comment

dorianjagusch (approved):
Good job!

As a remark, the Tournament/views.py file is barely readable to to the long line. Consider setting up a formatter for the backend. If no-wants to do that, I can. JLMK

Just for clarity, but I am pretty sure, I got it from the code and we discussed it: The players are not created with the tournament as discussed a few weeks ago, but instead are created with consecutive post requests after the POST tournamens/ to POST /tournaments/<tournament_id>/players/ right?

And lastly, both you and @merituulie will have to be careful with the merges since ya'll are touching some of the same files. I had some issues today, because old code got back into the main from our merges from yesterday, where git just reverted previous changes, if a newly merged branch didn't have those changes. I fell like this is not how git should behave, but just so you are aware.

sakarisalmi:
> As a remark, the Tournament/views.py file is barely readable to to the long line. Consider setting up a formatter for the backend. If no-wants to do that, I can. JLMK

I can definitely set up a formatter. Is Black good for you?
 
> Just for clarity, but I am pretty sure, I got it from the code and we discussed it: The players are not created with the tournament as discussed a few weeks ago, but instead are created with consecutive post requests after the POST tournamens/ to POST /tournaments/<tournament_id>/players/ right?

Yes

> And lastly, both you and @merituulie will have to be careful with the merges since ya'll are touching some of the same files. I had some issues today, because old code got back into the main from our merges from yesterday, where git just reverted previous changes, if a newly merged branch didn't have those changes. I fell like this is not how git should behave, but just so you are aware.

Yeah, I noticed this issue in the backend/urls.py file as well.

backend/Match/views.py
```diff
@@ -39,7 +40,7 @@ def get(self, request):
 class LaunchTestMatchView(APIView):
 	@method_decorator(must_be_authenticated)
 	def get(self, request):
-		token = MatchToken.objects.create_single_match_token(request.user, request.user)
+		token = MatchTokenManager.create_single_match_token(request.user, request.user)
 		match = MatchSetupManager.create_match_and_its_players(token)
 		if not match:
 			return Response({'error': 'Something went wrong when creating the match and players'}, status=status.HTTP_403_FORBIDDEN)
```
merituulie, line 46: the forbidden here is still wrong. I guess the 500 or Bad Gateway is best, if this is a db issue..... but not the best idea as previously discussed

backend/Tournament/views.py
```diff
@@ -51,34 +51,32 @@ def get(self, request, tournament_id):
...
-				return Response({'error': str(e)}, status=status.HTTP_403_FORBIDDEN)
-
-		elif tournament.state == TournamentState.IN_PROGRESS:
-			pass
+		try:
+			TournamentManager.setup.start_tournament(tournament)
+			serializer = TournamentSerializers.in_progress(tournament)
+			return Response(serializer.data, status=status.HTTP_200_OK)
+		except Exception as e:
+			return Response({'error': str(e)}, status=status.HTTP_403_FORBIDDEN)
```
merituulie, line 62: What is this forbidden thing????

backend/pong/consumers.py
```diff
@@ -68,39 +74,92 @@ async def send_positions_loop(self):
...
+                    user=winning_player.user
+                )
+
+                match.tournament.next_match += 1
+                match.tournament.save()
+
+                TournamentManager.in_progress.update_tournament_with_winning_tournament_player(winning_tournament_player)
+        
+        except Exception as e:
+            print(f"An error occurred while updating match results: {e}", file=sys.stderr)
```
merituulie, line 155: are we just ignoring the exception?

backend/Match/views.py
sakarisalmi: Let's fix this in issue #195

backend/Tournament/views.py
sakarisalmi: check comment above

backend/pong/consumers.py
sakarisalmi: added abort_tournament in exception

dorianjagusch:
> > As a remark, the Tournament/views.py file is barely readable to to the long line. Consider setting up a formatter for the backend. If no-wants to do that, I can. JLMK
> 
> I can definitely set up a formatter. Is Black good for you?
> 

I don't really care which as long as the style is consistent, spacing is as agreed and line lengths are ok. So feel free to go with black. Should be a separate PR though, cause once the code is formatted the diff is going to be crazy and substantial changes vs style changes are going to be hard filter through

backend/Tournament/views.py
```diff
@@ -51,34 +51,32 @@ def get(self, request, tournament_id):
 		return Response(serializer.data, status=status.HTTP_200_OK)
 
 	@method_decorator(must_be_authenticated)
-	@method_decorator(validate_tournament_request([TournamentState.LOBBY, TournamentState.IN_PROGRESS]))
+	@method_decorator(validate_tournament_request([TournamentState.LOBBY]))
 	def post(self, request, tournament_id):
```
dorianjagusch, line 55: am working with the API now and understand your thought process about the api and internal tournament management now. Instead of a post here,  I would suggest using a PATCH or PUT request to start the match. POST is for creating a new resource of the kind as given by the last url parameter (like POST tournaments/ -> new tournament, POST user/ -> new user etc). I would prefer PATCH, since PUT would require me to send all information of the tournament again.  With PATCH  one would typically send a body with the fields to update in the requests, but since we don't want the user to be fully in control (although, it could be nice to have the client send a request stating that the match has been aborted or so) of the tournament state, I suggest the client sends a PATCH /tournaments/<tournament_id> and the state is changed from LOBBY -> IN_PROGRESS and is rejected for when the tournament state is in any other and transitions to them as the games unfold. If we feel like it a PATCH in an IN_PROGRESS state could go to ABORTED.
 
 Not for this PR, but the POST doesn't seem right. A POST to tournaments/<id>/matches/ on the other hand, I could live with. The response could then contain the match id, on which the client could do a get or something. I haven't fully thought this through yet. Just thoughts

sakarisalmi: I've updated it now so that when the tournament.state == LOBBY, the tournament starts, and if the tournament.state == IN_PROGRESS, the tournament is aborted

sakarisalmi: also changed it to PATCH of course

dorianjagusch: So do we want to send from the frontend that a tournament is aborted? Like in between games  when no websocket is open and the client navigates away?

merituulie (approved):

dorianjagusch (approved):

PR #224, opened 2024-07-20
add scrolling to center canvas when going to pong page

dorianjagusch:
When the pong game canvas is created on the pong page, browser scrolls to center the canvas

## Changes:
- added scroll behaviour to pong page

## Caveats:
LMK if you want the instructions for the players to be closer to the bottom of the canvas, as the might not be visible depending on screen dimensions. I do know that #220 changes the playing field dimensions, so it might be void.


merituulie (approved):

azarSarikhani (approved):
great improvement!

PR #225, opened 2024-07-20
218 fix session management when user revisits but isnt logged in

### LINKED ISSUE #218: Fix session management when user revisits but isn't logged in
**Description**:
Everything is in the description

dorianjagusch:
User is logged out when the session is expired. Now that more pages (all actually via the navbar when loggedin) when a 401 is received, the user is redirected to the login page and  the localStorage is cleared.

## Changes:
- refactored navbar to have a setup and an update function
- refactored notifcations to be moved out of the navbar
- Modal notifications are red or green depending on the nature of the notification

## Next step:
When a user is on eg the friends page, the session expires and the user eg refreshes,  they should not only be forced to login again, but should then also be taken to the friends page
    EDIT: I played around with it for the past hour. It would require rewriting the router and that's more than I am willing to commit to rn in the interest of time. Could return to this, if other things are blocked

closes #218


azarSarikhani (approved):

merituulie:
In the interest of security I wouldnt go with the expired session -> no need to login again. Doesnt really even sound like something we should do!

merituulie (approved):

dorianjagusch:
>In the interest of security I wouldn't go with the expired session -> no need to login again. Doesnt really even sound like something we should do!

You mean that we shouldn't automatically relogin user's after the session expired? If yes I agree. I meant to redirect them when the session expired to make them login in, but then take them to the page they wanted to go to after logging in again. Like they were on the leaderboard page. the session expires after they went to the bathroom or something and the token expires. They click to go to the friends page, we detect that the session is expired and they are sent to the login page. They log in and are taken to the friends page instead to the dashboard (default after login). but as I said, that would require some redesign of the router. It may be nice for UX, but at this point, I don't think it's something we gotta do

sakarisalmi (approved):
Looks good!

Issue #229, opened 2024-07-20
Multiple consequitive games


dorianjagusch:
** Description **
Top run multiple consequitive games, either the backend musn't invalidate the token after a game or the frontend creeates a new token 
and saves the guest user's data (-> needs to be encrypted and the passphrase in turn needs to be sent from the backend)

sakarisalmi:
We can definitely set it up so that it is possible to play multiple games consecutively, but this isn't a mandatory requirement for the project, correct? We could definitely get away with forcing the host user to go through the setting up a new match again.

dorianjagusch:
As discussed, we will not allow users to continue playing, but they will have to reauthenticate the other player

PR #230, opened 2024-07-20
227 implement proper match communication

dorianjagusch:
Page flow works from login in to a full pong game

## Changes:
- Adjusted styles for playerinfoField
- Made accept decline modal take callback as argument
- Adjust url for tokens/authentication

## Caveats:
- Can't be used with the AI yet. I am very confused what the url/body should look like
- Cannot run multiple consecutive games yet with the same players #229


frontend/static/js/services/authenticationService.js
```diff
@@ -1,18 +1,24 @@
...
 export default class AuthenticationService extends ARequestService {
 	constructor() {
 		super();
 	}
 
-	async postRequest({username, password}) {
+	async postMatch({username, password}) {
 		return super.postRequest(
-			`${backendURL.authenticationURL}`,
+			`${backendURL.authenticationURL}match/ `, //${urlParams ? `/${urlParams}` : ''}`,
```
dorianjagusch, line 13: I left this in, as I will need it in the upcoming PRs (either today or tomorrow sunday)

frontend/static/js/services/authenticationService.js
```diff
@@ -1,18 +1,24 @@
...
 			JSON.stringify({
 				username: username,
 				password: password,
 			})
 		);
 	}
+
+	async postAiMatch() {
+		return super.postRequest(`${backendURL.authenticationURL}match/`);
```
dorianjagusch, line 22: I am really not sure what I need to do to get a token for the ai. There are the decorators on the view, which force me to user userdata in the body and th ai_opponent is a url parameter ??

azarSarikhani (dismissed):

frontend/static/js/services/authenticationService.js
merituulie: There is one token that is related to two users, and the db at least is okay the other user being null. If the decorator stops you from doing a POST request without the other user in the request body, than the decorator should be taken out of the server or changed to be okay with not getting tve other user if tje ai_opponent parameter is found. In general, I dont think you need another function to just pass the url oarameter but rather feed it into the service as is, like in FriendService GET all function. Check it out

frontend/static/js/services/authenticationService.js
```diff
@@ -0,0 +1,26 @@
...
+			`${backendURL.authenticationURL}match/ `, //${urlParams ? `/${urlParams}` : ''}`,
+			JSON.stringify({
+				username: username,
+				password: password,
+			})
+		);
+	}
+
+	async postAiMatch(URLParams) {
+		const url = new URL(`${backendURL.authenticationURL}match/`);
```
merituulie, line 22: meh, i dont think you need another method and a route for the AI match specifically.. otherwise, why not just have a special route for it in the server too? Just adding the parameter separately to either be empty or true / false with the basic Post request should be fine.

merituulie (approved):

frontend/static/js/services/authenticationService.js
dorianjagusch: it is just adding the parameter tokens/match//**?ai_opponent=true** instead of the body with username and password.  I need the postMatch to avoid more code duplication on the tournament flow, because the authentication modals, need to use different urls which need to be constructed in different ways. We can talk about it tomorrow in detail. My next PR might make the reason clearer as well. 

I am not happy with it either, but getting the token in conjunction with authenticating on one flow path, vs simply setting up players on the other, makes this the easier solution imo. But I am open to discuss :D

dorianjagusch: Also, if I were to do the modal flow again, I'd completely redesign it. I coded myself a little into a corner starting with the login and register modals from the very beginning

sakarisalmi (approved):
Looks good to me, but you know what that's worth.

sakarisalmi (approved):
Still nice!

sakarisalmi (approved):
Still nice!

PR #232, opened 2024-07-21
139 add stats app

azarSarikhani:
This PR will add a plot for user stats to be returned from the backend, and replaces returned plot from the backend for leaderboard with a json raw data
Screenshot 2024-07-21 at 14 52 42


backend/stats/friend_stats_table.py
```diff
@@ -0,0 +1,8 @@
+class FirendsTable:
```
sakarisalmi, line 1: FriendsTable :D

backend/stats/user_stats_table.py
```diff
@@ -0,0 +1,9 @@
+class UserTable:
+	def __init__(self, user, wins, losses, total_game_played, win_loss_ratio, wining_streak, position_in_leaderboard):
+		self.user = user
+		self.wins = wins
+		self.losses = losses
+		self.total_games_played = total_game_played
+		self.win_loss_ratioa = win_loss_ratio
```
sakarisalmi, line 7: win_loss_ratio*

backend/stats/user_stats_table.py
```diff
@@ -0,0 +1,9 @@
...
+		self.wining_streak = wining_streak
```
sakarisalmi, line 8: winning_streak*

backend/stats/views.py
```diff
@@ -7,49 +7,51 @@
...
 
-		return HttpResponse(plot_html, content_type='text/html')
-	
-class LeaderBoardView(APIView):
+class LeaderBoardView(APIView, UserTableMixin):
+	def __init__(self):
+		self.leader_table = LeaderBoardTable
+        
 	def get(self, request):
 		users_with_most_wins = User.objects.annotate(total_wins=Count('players', filter=Q(players__match_winner=True))).order_by('-total_wins')
```
sakarisalmi, line 46: I don't think you should have an __init__ in a view. How about you put almost everything you have in the get method into a mixin that returns Response(json_response_data, status=status.HTTP_200_OK)?
Then you could just do:
def get(self, request):
    return self.leader_table_response()

backend/stats/views.py
```diff
@@ -7,49 +7,51 @@
...
+from .user_stats_table import UserTable
+from .leader_board_table import LeaderBoardTable
+from .friend_stats_table import FirendsTable
 
+from .mixins import UserTableMixin
+from rest_framework.response import Response
 
-class StatsView(APIView):
+class StatsView(APIView, UserTableMixin):	
 	def get(self, request):
```
sakarisalmi, line 18: I don't know if you and Dorian have talked about whose job it is to show the stats, but it might look better if you return a json response to the frontend that then itself displays the stats data. Graphs and other visualisations of data we def need to do in the backend, of course.

sakarisalmi (commented):
Looks good, but fix the couple typos and check my other comments!

backend/stats/user_stats_table.py
```diff
@@ -0,0 +1,9 @@
+class UserTable:
+	def __init__(self, user, wins, losses, total_game_played, win_loss_ratio, wining_streak, position_in_leaderboard):
```
dorianjagusch, line 2: total_games_played

dorianjagusch: Winning_streak

backend/stats/views.py
dorianjagusch: No yeah, I'm with you @sakarisalmi
I guess something got lost in communication on Thursday. I do want the "raw" JSON data. Svg is fine for the graphs that show eg wins per day. If absolutely necessary, html elements are fine for play history of a player since they are only used in two places in the same way, but stats of a user is used in different ways all over the place. But even then, I don't see the point of sending html for the history of a player.

backend/stats/views.py
```diff
@@ -7,49 +7,51 @@
...
 from User.models import User
+from .user_stats_table import UserTable
+from .leader_board_table import LeaderBoardTable
+from .friend_stats_table import FirendsTable
 
+from .mixins import UserTableMixin
+from rest_framework.response import Response
 
-class StatsView(APIView):
+class StatsView(APIView, UserTableMixin):
```
dorianjagusch, line 17: This should probably be StatsDetailView mapped to the endpoint endpoint /stats/<user_id: int>
I don't think /stats and sending all the stats data of every user on the platform should be allowed.

backend/stats/views.py
```diff
@@ -7,49 +7,51 @@
...
+	
+		# Convert the Plotly figure to an SVG string
+		svg_str = fig.to_image(format="svg").decode("utf-8")
 
+		return HttpResponse(svg_str, content_type='image/svg+xml')
 
-		return HttpResponse(plot_html, content_type='text/html')
-	
-class LeaderBoardView(APIView):
+class LeaderBoardView(APIView, UserTableMixin):
```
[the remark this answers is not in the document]
dorianjagusch: And be in it's own app

backend/stats/friend_stats_table.py
dorianjagusch, line 1: should probably be in a history app together with play history eg. 
and could be HistoryFriendsTable, HistoryPlayTable etc. that gets hit on /histories/<user_id:int>/friends
and /histories/<user_id:int >/plays/, respectively.

dorianjagusch (changes requested):
Great job to get started on that. As mentioned and as we discussed on discord. I think the communication went awry, in terms of expected data. 
I also think this PR is trying to do to many more or less unrelated things. The different "tables" are working with similar data (mostly match/player data), but do very different things with it.
 
 The strength of REST is that we have a predictable interface where clients can easily reason about what to get from which endpoint by hitting certain endpoints that follow patterns. 
 
 Additionally, Django by nature is really modular and it's simply good practice to logically group code that belongs together. Due to us using REST, it would make sense to group the data by endpoint so those are together. That abstracts the database design away from the API and it and can be reasoned about together. Since, we don't have decent documentation on the API, it's already quite painful to dig through the code in order to find out what does what. When the code then is hidden somewhere where it doesn't belong, it becomes really painful...

So my suggestion for this PR is split it up. store the stuff on leaderboard, friend history etc in another branch or locally or whatever. Just implement the stats API to send bad request on GET /stats with a informative error message (provide a user id) or simply all stats for everyone, which I don't think we should allow. implement GET stats/<user_id:int> to return the stats for a single user in JSON format. if we want stats for eg by tournament etc (we don't currently have use for that, but if the group decides that or other things is something we want, sure), do  that. Basically everything that we currently are sure about, which is just by user id as far as I can tell. 

This makes it easier for you, so you don't have to juggle three different lines of thought at the same time and can focus on implementing one well and with clean code and maybe even performance considerations (which don't matter at this point, I'd rather get this project done). And it makes it easier for me, but more generally collaborators whose work is blocked on that front when non-working or many features are submitted rarely. 

I will design the APIs for stats/, /leaderboard and /history and discuss with Meri and Sakke tomorrow. we can then either have a call or simply take Tuesday to discuss

backend/stats/friend_stats_table.py
azarSarikhani: fixed!

backend/stats/mixins.py
```diff
@@ -0,0 +1,55 @@
...
+			win_streak = max(win_streak, current_streak)
+
+		return win_streak
+
+		return win_streak
+	
+	def get_users_with_most_wins(self):
+		return  User.objects.annotate(total_wins=Count('players', filter=Q(players__match_winner=True))).order_by('-total_wins')
+
+	def get_position_in_leaderboard(self, user: User):
```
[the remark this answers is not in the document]
azarSarikhani: In my opinion it is. I can add alphabetical order for those with same number of wins. If someone has only recently joined the website and has played 10 and has won 10 they might be a way better player than someone who has played 100 and only won 20 games but leaderboards don't show skills, I can add a row to the leaderboard that shows wining ratio, but this should not be a factor indicating who is on top of the leaderboard.

backend/stats/user_stats_table.py
azarSarikhani: fixed!

backend/stats/user_stats_table.py
azarSarikhani: fixed!

backend/stats/views.py
```diff
@@ -7,49 +7,51 @@
...
-                     ])
-
-        # Convert plot to HTML
-		plot_html = pio.to_html(fig, full_html=True)
+		user = request.user
+		wins = self.get_wins_count(user)
+		losses = self.get_losses_count(user)
+		total_game_played = self.get_total_games_played_count(user)
+		win_loss_ratio = self.get_win_loss_ratio(user)
+		wining_streak = self.get_win_streak_count(user)
```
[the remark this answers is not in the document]
azarSarikhani: fixed!

backend/stats/mixins.py
```diff
@@ -0,0 +1,55 @@
...
+	def get_total_games_played_count(self, user : User):
+		total_game_played_count = user.players.filter(match__state=MatchState.FINISHED).count()
+		return total_game_played_count
+	
+	def get_win_loss_ratio(self, user : User):
+		if self.get_total_games_played_count(user) == 0:
+			return 1.0
+		return self.get_wins_count(user) / self.get_total_games_played_count(user)
+	
+	def get_win_streak_count(self, user: User):
```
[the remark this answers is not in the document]
azarSarikhani: fixed!

backend/stats/user_stats_table.py
azarSarikhani: fixed!

backend/stats/views.py
```diff
@@ -7,49 +7,51 @@
...
-        </body>
-        </html>
-        """
+		self.leader_table.users = [user.username for user in users_with_most_wins]
+		self.leader_table.wins = [user.total_wins for user in users_with_most_wins]
+        
+		json_response_data = {
+            "users": self.leader_table.users,
+            "wins": self.leader_table.wins
+        }
```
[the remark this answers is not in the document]
azarSarikhani: Screenshot 2024-07-22 at 10 36 04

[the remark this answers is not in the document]
azarSarikhani: This is the returned json now. the first one on the list is the one with most wins and or the one higher on the alphabetised list. I can add winning ratio for each user if you wish.
And one more question do you want me to limit the list to first 10 top users or should I send the whole list as I am sending now?

backend/stats/views.py
azarSarikhani: @sakarisalmi done, I moved the whole thing to mixing. Do you think I need to have a separate app for leaderboard as well. I personally think it's an overkill, but if you think it's necessary then sure.

sakarisalmi: > Somebody discovered mixins xD

I have, and I'm in love 😍

sakarisalmi: > Do you think I need to have a separate app for leaderboard as well. I personally think it's an overkill.

No, but you can always use apps inside of apps, so if it makes it easier for you, you might want to do something like that.

backend/stats/views.py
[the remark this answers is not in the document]
azarSarikhani: I am uncertain about this name change. If the class LeaderBoardView was specifically designed to display a list of leaderboard entries I would have named it so, but it only returns one leaderboard json and not multiple ones. For reference look at class MatchView(APIView) it is named MatchView since it is not listing anything, but we have class FriendsListView(APIView) on the other hand. I changed the name, as requested but still puzzled about what listing means here.

backend/stats/mixins.py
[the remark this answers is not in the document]
dorianjagusch: As long as they have the same rank sure. 

regarding leaderboards not showing skills. I disagree with that. Often you have an actual high score which is a reflection of skill. There are scoring systems like in competitive chess that work on to reflect skill as well, that use the players history to adjust for scoring inflation and uncertainty in consistency when few games have played. 
like the [elo-rating system](https://en.wikipedia.org/wiki/Elo_rating_system#:~:text=The%20Elo%20rating%20system%20is,such%20as%20chess%20or%20esports.), [glicko-rating system](https://en.wikipedia.org/wiki/Glicko_rating_system) or [trueskill](https://en.wikipedia.org/wiki/TrueSkill). 
I think eg the elo is easy to implement. I am fine with simple win count in this PR, but would like to have a conversation about this, maybe best in rt communication, like during today's meeting or tomorrow in person :)

backend/stats/views.py
```diff
@@ -3,53 +3,30 @@
...
-		plot_html = pio.to_html(fig, full_html=True)
-
-
-		return HttpResponse(plot_html, content_type='text/html')
+		user = request.user
+		wins = self.get_wins_count(user)
+		losses = self.get_losses_count(user)
+		total_games_played = self.get_total_games_played_count(user)
+		win_loss_ratio = self.get_win_loss_ratio(user)
+		Winning_streak = self.get_win_streak_count(user)
```
dorianjagusch, line 17: lower case

backend/stats/views.py
```diff
@@ -43,15 +43,20 @@ def __init__(self):
...
-        
+
+		stats = [{"user": user, "wins": wins} for user, wins in zip(self.leader_table.users, self.leader_table.wins)]
+
 		json_response_data = {
-            "users": self.leader_table.users,
-            "wins": self.leader_table.wins
-        }
-        
+			"stats": stats
```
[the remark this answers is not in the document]
dorianjagusch: Doesn't need to be labelled at all, since I am expecting leaderboard items (or whatever we want to refer to them as as well) Also a rank/place would be good, but is not strictly necessary. The data should be self contained as much as possible

backend/stats/friend_stats_table.py
azarSarikhani: history has its own endpoint now

dorianjagusch (approved):

sakarisalmi (approved):
Nice!

dorianjagusch (approved):

PR #233, opened 2024-07-21
231 set up authenticate user mixin

sakarisalmi:
Created mixins for both authenticating and logging in a user. The former is an especially useful because we authenticate users in many different views. Planning on creating more mixins to make the views a lot more readable!


dorianjagusch (approved):
Would it make sense to wait with this PR until tournaments are through and apply it everywhere where it's needed?

merituulie (approved):

dorianjagusch (approved):

PR #235, opened 2024-07-22
Adjust ball speed after score if ball faster than original speed

merituulie:
Also added update to django version as its giving a security warning


sakarisalmi (approved):
Nice!

dorianjagusch (approved):
lgtm

PR #238, opened 2024-07-22
234 create more user app mixins

sakarisalmi:
Created a bunch of User app mixins.
Also formatted User app files to follow pep8 style.


merituulie:
Did you test all of the functions through frontend in this PR?

sakarisalmi:
> Did you test all of the functions through frontend in this PR?

I did, and all the User stuff works

backend/User/validators.py
```diff
@@ -2,16 +2,32 @@
 import os
 
 def validate_image(file):
-	valid_mime_types = ['image/jpeg', 'image/png', 'image/gif']
-	valid_file_extensions = ['.jpg', '.jpeg', '.png', '.gif']
-	max_file_size = 2097152
+    """
```
merituulie, line 5: mmeh, as said before, summariesa are quite unnecessary and get easily obsolete if not kept up to date...

backend/User/mixins.py
merituulie, line 1: I like this, picasso! Separating different database functionality to their own functions, very nice. This is not up to date though, fetch and pull from main! but Im not sure as we have a manager and a validator, what is the purpose of the mixin compared to those two... in anycase, dot really care at this point however meshed the responsibilities are but in general something to think about.

merituulie:
This PR has some conflicts. Is it still under construction or can it be re-reviewed already?

azarSarikhani (approved):

sakarisalmi:
Everything is now in order with this PR. I've tested all the stuff where the mixins are used and everything works like it should.

dorianjagusch (approved):

PR #241, opened 2024-07-22
fixed line 38 so that it calls MatchTokenManager and not MatchToken.o…

sakarisalmi:
…bjects


sakarisalmi:
The problem was the following: I delinked MatchTokenManager from MatchToken by removing from MatchToken model the line
objects = MatchTokenManager()

But in one of the views I was still calling a MatchTokenManager method with
... MatchToken.objects.FUNCTION ...
when it should have been
... MatchTokenManager.FUNCTION ...

that's all

dorianjagusch (approved):

merituulie (approved):

PR #242, opened 2024-07-22
196 close websocket on the backend

merituulie:
Fixed the AI opponent & token flow, close the websocket and return after the game is over.


backend/Match/views.py
```diff
@@ -15,7 +15,7 @@ class MatchView(APIView):
     @method_decorator(must_be_authenticated)
     def get(self, request):
         token_str = request.query_params.get('token')
-        if not token_str:
+        if not token_str or token_str == 'null':
```
dorianjagusch, line 18: Does python treat the null as true?

backend/pong/consumers.py
```diff
@@ -47,16 +45,18 @@ async def connect(self):
...
+        await self.close()
 
     async def receive(self, text_data):
         key_press = text_data.strip()
         self.game.update_positions(key_press)
 
         if self.game.game_stats.game_over == True:
             await self.send_positions()
-            await self.save_match_results()
+            await self.save_match_final_results()
```
dorianjagusch, line 57: So this is supposed to be back to final result?

dorianjagusch (approved):

backend/pong/consumers.py
sakarisalmi: There's some git shenanigans going on.

backend/pong/consumers.py
```diff
@@ -1,19 +1,17 @@
...
 import json
 import asyncio
+from asgiref.sync import sync_to_async
 from channels.generic.websocket import AsyncWebsocketConsumer
 from .game import PongStatus
 from .constants import *
 from channels.db import database_sync_to_async
 from Tokens.models import MatchToken
 from Match.models import Match
-from Match.matchState import MatchState
```
sakarisalmi, line 11: The imports are necessary. I recommend copy-pasting the stuff you've added in this pr to a pulled file from main. Some of the changes in this file tells me that something went wrong with git or something of the sorts.

backend/Match/views.py
merituulie: nope, this is a hack, the frontend given it does not find the token will send a string 'null' to the back

backend/pong/consumers.py
merituulie: There is some weirdness with the commits to this file for sure.... has there been a sqash merge or something?

backend/Match/views.py
dorianjagusch: ohhhh

backend/pong/consumers.py
dorianjagusch: Still shouldn't be possible at least not squashing PRs onto the main. What people do on their branches is up to them (so maybe?)

dorianjagusch (approved):

sakarisalmi (approved):
Nice, everything looks good!

Issue #237, opened 2024-07-22
Unify trailing blackslashes in urls


dorianjagusch:
**Description**:
trailing backslash handling is a little all over the place

PR #245, opened 2024-07-22
239 format backendbackend dir

sakarisalmi:
Tidying up different parts of the backend to get the project evaluation ready. Expect similar PRs of different backend apps soon.


azarSarikhani (approved):

backend/backend/urls.py
```diff
@@ -2,89 +2,106 @@
...
-	path('tournaments/<int:tournament_id>/matches/', TournamentMatchListView.as_view()),
-	path('tournaments/<int:tournament_id>/matches/<int:tournament_match_id>', TournamentMatchDetailView.as_view()),
-
-	# Stats Views
-	path('stats/', StatsView.as_view()),
-	path('leaderboard/', LeaderBoardView.as_view()),
+    path('admin/', admin.site.urls),
+
+    # User views
+    path('users/', UserListView.as_view(), name='user-list'),
```
merituulie, line 73: the name=user-list would be for use IF we used templates but as we dont, Im not sure we should add them here. Checking the documentation might help. But if it doesnt break anything than fine..

merituulie: just linking this issue here as itll be done in this one too, nice! https://github.com/dorianjagusch/ft_transcendence/issues/237

dorianjagusch: Will deal with it. Name is indeed not necessary

dorianjagusch:
Would be best,, if @sakarisalmi removed the name tags (tbf they don't hurt here either)

backend/backend/urls.py
sakarisalmi: I removed the name stuff!

PR #246, opened 2024-07-22
164 create tournament frontend

dorianjagusch:
Tournaments can now be created from the frontend and the first game can be played. 

** BIG CAVEAT ** 
CAREFUL, at the moment a failed login during player authentication, forces you to login and start again :sweat_smile: 

This overall worse than it is,  a lot of stuff is just spacing that my formatter did on the backend after the agreed format from months ago wasn't stuck to while I was digging through the code... 

Aaaaaanyway:

## Changes: 
- Integrated TournamentService into NumberOfPlayersModal, SelectPlayersModal and SummaryModal
- Refactored the play page and SelectPlayersModal
- Apply eventhandlers via the caller for the Summary modal instead of in itself
- Added pong/tournaments/<tournament_id>/matches/<match_id> endpoint on frontend (mostly for information transfer between pages)
- Completed flow from creating the tournament to playing it

## Caveats
- Currently only one game can be played as the transition to the next game is not there (no idea what request to what endpoint with what data)


backend/User/views.py
```diff
@@ -146,8 +147,6 @@ def get(self, request, user_id):
 				image_data = image_file.read()
 				encoded_image = base64.b64encode(image_data).decode('utf-8')
 				return Response({'image': encoded_image}, status=status.HTTP_200_OK)
-		except ProfilePicture.DoesNotExist:
```
merituulie, line 149: Whats up? We throw an error if the photo is not found or does the encoded picture just return as empty image?

frontend/static/js/services/tournamentService.js
```diff
@@ -1,33 +1,58 @@
...
+				password: tournamentData.password,
+			})
+		);
 		return response;
 	}
 
-	async deleteRequest(){
-		throw new NotImplementedError("DELETE not implemented for TournamentService");
+	async patchRequest(context) {
+		return super.patchRequest(backendURL.tournamentURL, context.tournamentId);
```
merituulie, line 48: mmeh, the patch body is supposed to look like:
{
  'op' : 'replace',
  'path' : 'tournament.status',
  'value': 'lobby'
}

to actually be a valid expected patch call, but we dont have time to do the right thing..

merituulie (approved):

dorianjagusch:
Sorry for the revert commits. I accidentally continued working on this branch although I had created another one. Just reverted the commits  to its old state where Meri commented.

sakarisalmi (approved):
Great, we're so close!

backend/User/views.py
```diff
@@ -135,8 +136,8 @@ def post(self, request, user_id):
 
 	@method_decorator(must_be_authenticated)
 	def get(self, request, user_id):
+		user = get_object_or_404(User, pk=user_id)
```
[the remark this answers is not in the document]
dorianjagusch: We might actually not want to use it to get informative messages. I did a quick google search and there seems to be no good way to add a message to the 404 from this

backend/User/views.py
merituulie: so if the image_data is not there and we feed null to the base64 encoding, its fine, even with the decode question? to me this change looks to be unnecessary.

merituulie: Maybe just make it explicit that we return empty image + status 200 thingie if profile picture is not found, thanks!

merituulie (approved):

merituulie (approved):

PR #257, opened 2024-07-23
253 unify error messages from backend to message error message statuscode

merituulie:


backend/User/views.py
```diff
@@ -131,25 +131,25 @@ def post(self, request, user_id):
...
-				return Response({'image': ''}, status=status.HTTP_200_OK)
+				return Response({"image": ""}, status=status.HTTP_200_OK)
 
 			image_path = profile_picture.picture.path
 			with open(image_path, "rb") as image_file:
 				image_data = image_file.read()
 				encoded_image = base64.b64encode(image_data).decode('utf-8')
-				return Response({'image': encoded_image}, status=status.HTTP_200_OK)
+				return Response({"image": encoded_image}, status=status.HTTP_200_OK)
 		except ProfilePicture.DoesNotExist:
```
dorianjagusch, line 149: So how do we want to deal with this, as we begun discussing this in #246 ?
The ProfilePicture.DoesNotExit exception is not going to be thrown, but 2 others could be IOError and base64.binascii.Error ?

dorianjagusch (approved):

azarSarikhani (approved):

PR #262, opened 2024-07-24
Fix collision detection, fix AI paddle movement

merituulie:
A small fix for the AI and the ball angle to make sure it bounces nicely on the sides of the paddle too. Speed tolerance still doesnt quite work in high speed but I guess it doesnt mattr that much for this project.

### DIFF
--- a/backend/pong/game_logic.py
+++ b/backend/pong/game_logic.py
@@ -65,12 +65,13 @@ class PongGame:
         ball_y = pong_stat.ball.y
         impact_point = ball_y - player_y
         normalized_impact_point = impact_point / HALF_PLAYER_HEIGHT
-        new_angle = normalized_impact_point * MAX_BOUNCE_ANGLE_RAD * 2
+        bounce_angle = normalized_impact_point * MAX_BOUNCE_ANGLE_RAD
 
-        if self.half_pi <= pong_stat.ball.angle < 3 * self.half_pi:
-            pong_stat.ball.angle = (new_angle)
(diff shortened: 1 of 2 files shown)


backend/pong/game_logic.py
```diff
@@ -65,12 +65,13 @@ def handle_player_collision(self, pong_stat, player_y):
         ball_y = pong_stat.ball.y
         impact_point = ball_y - player_y
         normalized_impact_point = impact_point / HALF_PLAYER_HEIGHT
-        new_angle = normalized_impact_point * MAX_BOUNCE_ANGLE_RAD * 2
+        bounce_angle = normalized_impact_point * MAX_BOUNCE_ANGLE_RAD
 
-        if self.half_pi <= pong_stat.ball.angle < 3 * self.half_pi:
-            pong_stat.ball.angle = (new_angle)
+        if pong_stat.ball.x < PLAYGROUND_WIDTH / 2:
+            pong_stat.ball.angle = (2 * math.pi - bounce_angle) % (2 * math.pi)
```
dorianjagusch, line 71: the simple modulo operator doesn't work well for floats due to precision issues (eg. 3.5 % 0.1 = 0) but `3.5 % 0.1 = 0.09999999999999981`, but it might be fine here

dorianjagusch (approved):

backend/pong/game_logic.py
dorianjagusch: We could also check whether $0 < angle < 2\pi$ is true and add or subtract $2\pi$ if the angle is bellow the range or above the range, if precision matters :grimacing:

PR #264, opened 2024-07-24
Fix authservice to matchService

dorianjagusch:
Left over authentication service removed and switched out with match service


azarSarikhani (approved):

sakarisalmi (approved):
radical change I see

PR #270, opened 2024-07-24
Edit username frontend

merituulie:
Ugly new modal for editing the profile, either username or password or both. There is still a small bug, that I believe has to do with the modal / form not updating / being created a new but couldn't figure it out at this point, would like to fix it on the next PR - unless Dorian has a great idea for it. Here's a video of the functionality and about the bug in the second case:

https://github.com/user-attachments/assets/89f08f3c-db3b-4e35-997a-fa2697bb2cc1


azarSarikhani (approved):

dorianjagusch (approved):
I don't follow. What's the bug exactly?

I will investigate the branch properly tomorrow, but am taking it easy today before I collapse 😅

PR #271, opened 2024-07-24
266-change-tournament-player-name-to-deleted-user

sakarisalmi:
created Tournament mixin for changing a user's tournamentplayer display_names to 'deleted_user'.
used that mixin in User's DeleteUserMixin.

tested, works.

Not related to this pr or issue, but I'm not really happy with my naming conventions in some of the User app's mixins and views; For example, because something like GetUserMixin's get_user function can return either a user or a response, I don't want the code to have
`user = self.get_user(user_id)`
because the result could not be the user, and in that case it would be a bit strange to have something like
```
if not is instance(user, User):
        return user
```

so because of this I usually use
`result = self.get_user(user_id)`
but then for the rest of the mixin or view I have stuff like this
```
result.username = f"deleted_user_{user_id + 42}"
        result.set_password(get_random_string(length=30))
        result.is_active = False
        result.is_staff = False
        result.is_superuser = False
        result.insertTS = None
        result.last_login = None
        result.is_online = False
        result.save()
```
which is also strange.

Should I just write the code like this
```
result = self.get_user(user_id)
if not isinstance(result, User):
        response = result
        return response
user = result
#continue with function
```
even if it adds a couple of unnecessary lines? Just some general thoughts I'd appreciate some comments on...

### DIFF
--- a/backend/User/mixins.py
+++ b/backend/User/mixins.py
@@ -11,6 +11,7 @@ import base64
 from .models import User, ProfilePicture
 from .validators import validate_image
 from .serializers import UserInputSerializer, UserOutputSerializer
+from Tournament.mixins import ChangeDeletedUserTournamentNamesMixin
 
 class GetAllUsersMixin:
     """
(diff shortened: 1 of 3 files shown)


backend/User/mixins.py
```diff
@@ -111,29 +112,32 @@ def logout_user(self, request: Request) -> Response:
...
+        result = get_object_or_404(User, pk=user_id)
+        if not isinstance(result, User):
+                return result
+
+        result.username = f"deleted_user_{user_id + 42}"
+        result.set_password(get_random_string(length=30))
+        result.is_active = False
+        result.is_staff = False
+        result.is_superuser = False
+        result.insertTS = None
```
dorianjagusch, line 133: Why 🐫 instead of 🐍?

dorianjagusch (approved):

backend/User/mixins.py
sakarisalmi: I remember that that field had been changed ages ago. I'll fix it now.

merituulie:
Good thoughts on the naming! At this point, unless we have some extra time, we can change that but for the lifetime of this project I dont see the need for the change at this point. But ofc one can always fork this project later and do it!

merituulie (approved):

PR #273, opened 2024-07-25
Add styles, fix password read

merituulie:
Fixed the bug from the editing user. If the user updates the password, we log them out so they have to login again. If the user just changes the username, we just save the new username and update the UI. Added some more information to the modal and the notifications depending on if we update just the username, and/or the password.

https://github.com/user-attachments/assets/091dec1c-c73f-4de9-bf41-53812e147188


sakarisalmi (approved):
Nice!

dorianjagusch (approved):

PR #282, opened 2024-07-25
281 create player views

sakarisalmi:
I've created three different player views for retrieving the serialized data of players.
get all players of all matches: /players/
get a specific player by player_id: /players/<int:player_id>
get players of a specific match: /matches/<int:match_id>/players/

tested, works

### DIFF
--- a/backend/backend/urls.py
+++ b/backend/backend/urls.py
@@ -35,6 +35,10 @@ from Tokens.views import SingleMatchGuestTokenView
 from Match.views import MatchView, \
 							MatchHistory
 
+from Player.views import PlayerListView, \
+							PlayerDetailView, \
+							MatchPlayerListView
+
 from Tournament.views import TournamentListView, \
 								TournamentDetailView, \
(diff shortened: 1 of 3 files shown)


backend/backend/urls.py
```diff
@@ -68,6 +72,11 @@
 	path('match/', MatchView.as_view()),
 	path('matches/', MatchHistory.as_view()),
 
+	# Player views
+	path('players/', PlayerListView.as_view()),
+	path('players/<int:player_id>', PlayerDetailView.as_view()),
```
dorianjagusch, line 77: What happened to the PR that was fixing those?

dorianjagusch, line 77: What happened to the PR that was fixing those?

dorianjagusch (approved):
Noice

merituulie (approved):

backend/backend/urls.py
sakarisalmi: what do you mean? Fix what?

PR #284, opened 2024-07-25
Add player match score graph

azarSarikhani:
This Pull Request will add match graph at 'users/<int:user_id>/matches_graph' url
In the image bellow the url is different, please disregard that. Comment on style of the bar, modifications could easily get added. Bar is shown as green if player has won that game and red if they haven't.
Screenshot 2024-07-25 at 18 12 24

### DIFF
--- a/backend/backend/urls.py
+++ b/backend/backend/urls.py
@@ -43,7 +43,8 @@ from Tournament.views import TournamentListView, \
 								TournamentMatchDetailView
 
 from stats.views import StatsView, \
-						StatsGraphView
+						StatsGraphView, \
+						MatchScoreGraphView
 
 from leaderboard.views import LeaderboardListView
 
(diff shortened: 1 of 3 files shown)


backend/backend/urls.py
```diff
@@ -79,7 +80,8 @@
 	# Stats Views
 	path('users/<int:user_id>/stats/', StatsView.as_view()),
 	path('leaderboard/', LeaderboardListView.as_view()),
-	path('users/<int:user_id>/graphs/', StatsGraphView.as_view())
+	path('users/<int:user_id>/graphs/', StatsGraphView.as_view()),
```
dorianjagusch, line 83: Technically, neither a bar nor a pie chart are graphs

dorianjagusch (approved):

backend/backend/urls.py
azarSarikhani: should I rename them to chart?

merituulie: maybe just something like users/user_id/last_played_matches_score?

merituulie (approved):

PR #285, opened 2024-07-25
267 deleted users photo deletion

hussojo:
Changed updating the profile picture so that old picture is deleted when updating to new one. Deleting user picture and the row from database when user gets deleted. Updated html.

### DIFF
--- a/backend/User/mixins.py
+++ b/backend/User/mixins.py
@@ -110,7 +110,8 @@ class LogoutUserMixin:
         logout(request)
         return Response({"message": "User logged out"}, status=status.HTTP_200_OK)
 
-
+import os
+from django.conf import settings
 class DeleteUserMixin:
 	"""
 	Mixin to delete a user by ID.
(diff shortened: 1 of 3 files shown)


sakarisalmi (approved):
Nice!

backend/User/mixins.py
```diff
@@ -149,10 +149,14 @@ def delete_user(self, request: Request, user_id: int) -> Response:
         result.last_login = None
         result.is_online = False
         result.save()
+        profile_picture = ProfilePicture.objects.filter(user=result).first()
+        if profile_picture:
+            profile_picture.delete_profile_picture()
+            profile_picture.delete()
```
merituulie, line 155: why are there two deletes? wouldnt it be enough to just delete the db row

frontend/privacypolicy.html
```diff
@@ -72,7 +72,7 @@ <h2>3. Your Data Protection Rights</h2>
         <li><strong>The right to erasure</strong> – You have the right to request that we erase your personal data, under certain conditions.</li>
     </ul>
 
-    <p>If you make a request, we have one month to respond to you. If you would like to exercise any of these rights, please contact us at our provided contact information.</p>
+    <p>If you make a request, we have one month to respond to you. If you would like to exercise any of these rights, please contact us at poxelor275@vasomly.com.</p>
```
merituulie, line 75: :D

merituulie (approved):

PR #291, opened 2024-07-25
274 winner page

dorianjagusch:
Tournament flow until wining page.

## Changes:
- flow of tournament completed (only display of information missing)
- Added winner page 
- refactored backend url creation to use parts of it else where (here: 'localhost:8080' to be used for regex for the websocket)
- Distinguish between previewPage and winnerPage
- Renamed matchService to authentication Service -> releases name for /matches endpoints


azarSarikhani (approved):

frontend/static/js/constants.js
```diff
@@ -1,4 +1,6 @@
-const baseURL = 'http://localhost:8080/';
+const protocol = 'http://';
```
merituulie, line 1: nice

merituulie (approved):

PR #294, opened 2024-07-26
292 add more test to the new rest api tests postman collection

sakarisalmi:
This PR adds match and tournament tests to the (new) postman collection and environment.

Some other changes:
- prevent a single match token from creating multiple matches with a new flag.
- removed guest_user from the Response when creating AI match token
- improved create_tournament_and_tournament_player_for_host function
  - use get for validated data
  - delete tournament if creation of host tournament player fails
- added back exception string to PATCH TournamentDetailView Response message (tells that there aren't enough players to start tournament)

### DIFF
--- a/backend/Match/views.py
+++ b/backend/Match/views.py
@@ -15,22 +15,27 @@ class MatchView(APIView):
     def get(self, request):
         token_str = request.query_params.get('token')
         if not token_str or token_str == 'null':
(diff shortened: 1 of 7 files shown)


dorianjagusch (approved):
Noice, but just putting it out there. Is that the best use of the time we still have. We do only have the next two weeks and there might be plenty of bugs hiding, we don't know about

merituulie (approved):

PR #297, opened 2024-07-26
295-send-pong-names-from-backend

### LINKED ISSUE #203: Pass player names to the game
**Description**:
 Currently we have the placeholder names in the game. Should be the user/custom names of the actual players

### LINKED ISSUE #295: display names in pong game
Display usernames in normal matches (also handle AI).
Display tournament player display names if tournament match.

sakarisalmi:
Added a function for retrieving appropriate names at the start of the pong game. The names will be sent in the beginning with the constants (players.left_name and players.right_name).

closes #295 
closes #203

### DIFF
--- a/backend/pong/game.py
+++ b/backend/pong/game.py
@@ -89,11 +89,13 @@ class PongStatus:
 
         return ball_y
 
-    def get_consts(self):
+    def get_consts(self, left_name: str, right_name: str) -> dict[str, any]:
         game_consts = {
             'players': {
                 'width': 2 * HALF_PLAYER_WIDTH,
                 'height': 2 * HALF_PLAYER_HEIGHT,
+                'left_name': left_name,
+                'right_name': right_name
             },
             'ball': {
                 'size': 2 * HALF_BALL_SIZE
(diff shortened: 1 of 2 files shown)


dorianjagusch (approved):

merituulie (approved):

backend/pong/game.py
```diff
@@ -89,11 +89,13 @@ def calculate_ai_steps(self):
 
         return ball_y
 
-    def get_consts(self):
+    def get_consts(self, left_name: str, right_name: str) -> dict[str, any]:
         game_consts = {
             'players': {
                 'width': 2 * HALF_PLAYER_WIDTH,
                 'height': 2 * HALF_PLAYER_HEIGHT,
```
merituulie, line 96: we can do it like this, but to me makes more sense to use the selected display names and the frontend just gets them once when creating the play view.

azarSarikhani (approved):

backend/pong/game.py
sakarisalmi: You can change it if you want, I just thought it would be a lot easier for the frontend that it get's the names from at the start of the match so that it doesn't require a separate request.

PR #300, opened 2024-07-26
Reset players when one scores

merituulie:
We now reset player position when one scores, as the field is quite large and the first ball is hard to predict.


azarSarikhani (approved):

dorianjagusch (approved):

dorianjagusch:
Whoohoo 300 PRs

PR #309, opened 2024-07-29
299 redesign and complete dashboard

dorianjagusch:
New dash board page uses the matches API to created history list. 
Something broke for the scroll container, which will be treated separately.
clicking on the match, sends you to the profile page of the opponent. 

## To Come:
Fixing of the scroll component
Integration of the images from the backend


frontend/static/js/components/profileComponents/profilePlayHistory.js
```diff
@@ -1,10 +1,16 @@
-const profilePlayHistory = (history) => {
+// const history = {
```
merituulie, line 1: ?

merituulie (approved):

merituulie (approved):

PR #311, opened 2024-07-29
310 fix win page

### LINKED ISSUE #310: fix win page
**Description**:
Win and match pages, don't get the current data of a user

dorianjagusch:
Win- match- and preview pages display the stats of the participants
End of game shows winner name

closes #310


sakarisalmi (approved):
I'll test this in the frontend, but looks pretty good from what I can tell!

PR #315, opened 2024-07-29
312 fix images on frontend

dorianjagusch:
## Changes: 
- Add images to frontend
- Fix deleted user handling on frontend
- start of fixing history component (it still overflows tho)

### DIFF
--- a/frontend/static/js/services/ARequestService.js
+++ b/frontend/static/js/services/ARequestService.js
@@ -138,7 +138,7 @@ class ARequestService {
 			method: 'DELETE',
 			headers: {
 				'X-CSRFToken': getCookie('csrftoken'),
-				'Content-Type': 'application/json',
+				'Content-Type': 'image/svg+xml',
 			},
 			credentials: 'include',
 		});
--- a/frontend/static/js/services/ARequestService.js
+++ b/frontend/static/js/services/ARequestService.js
@@ -138,7 +138,6 @@ class ARequestService {
 			method: 'DELETE',
 			headers: {
 				'X-CSRFToken': getCookie('csrftoken'),
-				'Content-Type': 'image/svg+xml',
 			},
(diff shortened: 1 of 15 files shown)


azarSarikhani (approved):

sakarisalmi (approved):

frontend/static/js/services/ARequestService.js
```diff
@@ -138,7 +138,7 @@ class ARequestService {
 			method: 'DELETE',
 			headers: {
 				'X-CSRFToken': getCookie('csrftoken'),
-				'Content-Type': 'application/json',
```
merituulie, line 141: I dont think this is a good idea, the rest of the deletions will be json wont they?

merituulie: In general, deleteRequest doesnt really take a body, just an id of the resource we want to delete. how about an overload?

frontend/static/js/services/statsService.js
```diff
@@ -11,15 +11,23 @@ export default class StatsService extends ARequestService {
...
+				},
+				credentials: 'include',
+			});
+			if (!response.ok) {
+				throw new Error('Network response was not ok');
+			}
+			const svgText = await response.text();
+			return svgText;
+		} catch (error) {
+			console.error("Couldn't fetch data", error);
```
merituulie, line 30: notify?

frontend/static/js/views/dashboardView.js
```diff
@@ -64,15 +53,14 @@ export default class extends AView {
 				return {
 					match_id: match.id,
 					opponent: opponent.username,
-					opponentId: opponent.id,
```
merituulie, line 67: fetch and pull!

azarSarikhani (approved):

PR #342, opened 2024-07-31
lowered max image size

dorianjagusch:
This caused some timing issues, making the logout not happen fast enough

### DIFF
--- a/backend/User/validators.py
+++ b/backend/User/validators.py
@@ -5,7 +5,7 @@ import re
 def validate_image(file):
 	valid_mime_types = ['image/jpeg', 'image/jpg', 'image/png']
 	valid_file_extensions = ['.jpg', '.jpeg', '.png']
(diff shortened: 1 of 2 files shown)


backend/User/validators.py
```diff
@@ -5,7 +5,7 @@
 def validate_image(file):
 	valid_mime_types = ['image/jpeg', 'image/jpg', 'image/png']
 	valid_file_extensions = ['.jpg', '.jpeg', '.png']
-	max_file_size = 2097152  # 2MB
+	max_file_size = 1048576
```
merituulie, line 8: is this fixed from the frontend too?

merituulie: and settings.py

dorianjagusch: does the frontend doesn't check and just sends to the backend whatever it gets

sakarisalmi (approved):

PR #343, opened 2024-07-31
Remove extra files

dorianjagusch:
removing some files that have fallen out of use


sakarisalmi (approved):

sakarisalmi (approved):

merituulie (approved):

backend/User/validators.py
```diff
@@ -5,14 +5,14 @@
 def validate_image(file):
 	valid_mime_types = ['image/jpeg', 'image/jpg', 'image/png']
 	valid_file_extensions = ['.jpg', '.jpeg', '.png']
-	max_file_size = 2097152  # 2MB
+	max_file_size = 1048576
```
merituulie, line 8: the frontend also checks for size, so it should match

merituulie: validateFile in frontend

merituulie (approved):

sakarisalmi (approved):

PR #345, opened 2024-07-31
Ssalmi fix pong bug

sakarisalmi:
The bug that caused a player to play against themself in a tournament is fixed.
The problem was that if you press a button when the match finishes, the receive function would call the save_match_results function, which would mean that save_match_results would get called twice. This would mean that the function update_tournament_data_with_match_results would get called twice, which would cause the winner of the match to be assigned twice to the next tournament match.

With these changes this specific bug has been fixed, but there is still some errors going on in the backend container (autobahn.exception.Disconnected: Attempt to send on a closed protocol) which need to be fixed. But still!

There's also some other small things in this PR by @merituulie.


backend/pong/consumers.py
```diff
@@ -2,6 +2,7 @@
 from django.db import transaction
 import json
 import asyncio
+import sys
```
merituulie, line 5: lets remove this!

dorianjagusch (approved):

backend/pong/consumers.py
sakarisalmi: ok

dorianjagusch (approved):

PR #359, opened 2024-08-01
add in a column for score zero

azarSarikhani:
This PR adds in a small bar for zero scores
Earlier it was a bit confusing as it would show a blank space between bars even though there wasn't a zero point match. Now we avoid this problem.
Screenshot 2024-08-01 at 15 05 08


sakarisalmi (approved):
Nice!

PR #362, opened 2024-08-19
361 refactor backend into a more cohesive entity

### LINKED ISSUE #361: Refactor backend into a more cohesive entity
I'd like to make the backend into a more cohesive (and generally nicer) entity. This has to do both with its structure and formatting. I think it would be cleaner if everything that is done in the different managers would be moved into mixins that will then be called in the actual view functions. I'll refactor the project app by app.

If anyone else does not want these late additions to the project, please tell me, as I can always simply fork the project and work on refactoring in a separate repo. I do think that these additions are useful for everyone in the project though. The goal with the refactoring is not to change anything in how the backend functions, it's just make the backend generally cleaner and more understandable.

sakarisalmi:
In this PR I've created mixins for Friend, Tokens, Match, Player and Tournament apps. Also made slight changes to the User app, but otherwise it's basically the same as earlier.

I've removed all the managers whose functions I've moved into the mixins. The only manager I didn't remove is the TournamentManager, but even that I've simplified: now there is only one TournamentManager instead of the multiple separate ones that were earlier there earlier.

I've also done some slight formatting to the apps mentioned above. Also removed useless imports.

I've tested all the apps that I've touched with Postman and everything works like it should!


merituulie:
Nice, but I'm expecting you also test everything from the UI before I can accept this PR. Also in the future one can just create a fork from this repo and further the development there!

merituulie (changes requested):
Awaiting your status update for testing everything in the UI.

sakarisalmi:
> Awaiting your status update for testing everything in the UI.

Ok, I tested it on the frontend with the production stuff, and I did forget to make small changes in the pong/consumers.py which caused the tournament not to update after a match. Now it is fixed.

Everything else works like it should when testing with the UI!

merituulie (dismissed):
Approving! Nice job, but just making sure you tested **all** of the features used by the client in affected endpoints of the server?

merituulie:
Like in general you can think of it like this: backend project is the base for this project, the concrete and the bearing structures and the pipes for the house.  The front end is the different floors and the stairs, the ceiling, walls and paint over it. If you just rip out one of the bearing sturctures to maintain the house, there is going to be a hole in the wall, or the whole house might crumble, right?

The front end is based on and dependent on the backend so changing anything on the backend has to tested also in the frontend. This is why Im asking you to test the changes also in the frontend always.

dorianjagusch:
So what's the status on this rn?

sakarisalmi:
Oh, sorry for not touching this in a while. I did test all the features in it the frontend, and it did work, but I'd like for you guys to test it as well just in case.

Otherwise I think it is ready to merge!

merituulie:
I'd suggest you create a fork from this repository and implement further changes there :)

dorianjagusch:
Yeah, I don't think it is feasible to expect anyone to do any work on this anymore. I WILL, however, merge this. If something is in fact broken. Peeps can revert to the previous state in their own forks, if they haven't forked the repo yet.

